PR #11, opened 2020-08-14
Ensure nisystemlink-clients-python treats systemlink as an implicit namespace package

alexweav:

`systemlink` is a [PEP-420 style namespace package](https://www.python.org/dev/peps/pep-0420/). This mechanism is used by Python-based SystemLink services, allowing the service code to live under the `systemlink` namespace, but exist elsewhere on the module search path, sandboxed outside of `site-packages`. After installing `nisystemlink-clients-python`, an `__init__.py` is dropped in the `site-packages` portion of the namespace, causing Python to ignore other packages within the namespace, elsewhere in the search path.

As [documented](https://packaging.python.org/guides/packaging-namespace-packages/):
> It is extremely important that every distribution that uses the namespace package omits the __init__.py or uses a pkgutil-style __init__.py. If any distribution does not, it will cause the namespace logic to fail and the other sub-packages will not be importable.

This PR causes the `__init__.py` in the `systemlink` namespace to be omitted from the produced .whl, restoring the PEP-420 behavior.

There are [numerous](https://github.com/python/mypy/issues/8944) [issues](https://github.com/python/mypy/issues/5759) when using mypy on namespace packages, so the `__init__.py` in the `systemlink` directory is not actually removed as part of this PR. As far as mypy is concerned, this is a regular package. The `setup.py` is updated to treat `systemlink` as a namespace package, and that `__init__.py` is not included in the .whl.

Manually inspected the generated .whl to ensure that the file was not present. Installed the generated package and verified that other `systemlink` modules elsewhere on the search path were importable. Installed the generated package by itself on a fresh system and verified that it is importable.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -22,7 +22,8 @@ jobs:
         python-version: ${{ matrix.python-version }}
     - name: Install dependencies
       run: |
-        python -m pip install --upgrade pip setuptools wheel
+        python -m pip install --upgrade pip
+        python -m pip install --upgrade setuptools wheel
         pip install tox tox-gh-actions
     - name: Run tox
       run: |
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -22,7 +22,7 @@ jobs:
         python-version: ${{ matrix.python-version }}
     - name: Install dependencies
       run: |
-        python -m pip install --upgrade pip
+        python -m pip install --upgrade pip setuptools wheel
         pip install tox tox-gh-actions
     - name: Run tox
       run: |
--- a/setup.py
+++ b/setup.py
@@ -56,10 +56,8 @@ setup(
         "events",
         'httpx;python_version>="3.6"',
         'requests;python_version<"3.6"',
-        "setuptools>=40.1.0",
         "typing-extensions",
     ],
-    setup_requires=["setuptools>=40.1.0"],
     tests_require=["pytest", "pytest-asyncio", "mypy"],
     classifiers=[
         "Development Status :: 5 - Production/Stable",
--- a/setup.py
+++ b/setup.py
@@ -56,6 +56,7 @@ setup(
         "events",
         'httpx;python_version>="3.6"',
         'requests;python_version<"3.6"',
+        "setuptools>=40.1.0",
         "typing-extensions",
     ],
     setup_requires=["setuptools>=40.1.0"],
--- a/setup.py
+++ b/setup.py
@@ -1,4 +1,4 @@
-from setuptools import find_packages, setup  # type: ignore
+from setuptools import find_namespace_packages, find_packages, setup  # type: ignore
 from setuptools.command.test import test as TestCommand  # type: ignore
 
 
@@ -15,6 +15,10 @@ class PyTest(TestCommand):
 
 
 pypi_name = "nisystemlink-clients"
+
+packages = find_namespace_packages(include=["systemlink.*"]) + find_packages(
+    exclude=["systemlink", "examples", "tests"]
+)
 
 
 def _get_version(name):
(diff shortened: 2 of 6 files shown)


setup.py
```diff
@@ -46,14 +50,15 @@ def _read_contents(file_to_read):
...
-    packages=find_packages(exclude=["examples", "tests"]),
+    packages=packages,
     install_requires=[
         'aenum;python_version<"3.6"',
         "events",
         'httpx;python_version>="3.6"',
         'requests;python_version<"3.6"',
         "typing-extensions",
     ],
+    setup_requires=["setuptools>=40.1.0"],
```
alexweav, line 61: This is discouraged in favor of a `pyproject.toml` value, but this repo does not have a `pyproject.toml`. If you would prefer that I add one, please let me know.

alexweav:
Interestingly, the change to `tox.ini` causes the `py35` build to try to run ALL FOUR build environments. It then tries to run `black` on py35, which is unsupported, so that build fails. My change doesn't seem to be related to the build matrix functionality at all, confusingly - currently looking into a fix.

DavidCurtiss:
> .tox create: D:\a\nisystemlink-clients-python\nisystemlink-clients-python\.tox\.tox

It looks like adding a setup_requires causes it to run tox inside of tox. And the internal tox doesn't obey the gh_actions. I dunno why it decides to run tox inside of tox, though.

alexweav:
It seems to be not the setup_requires, but rather the `requires` tag in the tox.ini. I've tried many combinations of `setup_requires`, `install_requires`, `pyproject.toml`, etc. Without the requires tag, the py35 build fails because setuptools is too old, and with it, we get the strange tox-within-tox behavior where the version constraints are ignored.

I suspect this might be a bug with `tox-gh-actions`, but I'm unsure. Perhaps there is a way to get the same parallel build behavior without needing that package? I can test stripping it out and see if that fixes anything.

alexweav:
When `tox` gets a `requires` entry that is not installed in the current system, it prefers not to alter the configuration of the system whatsoever. Instead it creates a shim `virtualenv` which satisfies all the `requires`, and runs the actual `tox` suite inside of that, meaning two nested venvs. However, `tox-gh-actions` does not expect this, and so the config that it injects (describing which tox env to run) is lost to the inner venv, causing the entire test suite to run as if it's running on a dev machine. It then tries to run `black`, which dies on py35.

On top of that, the FIRST thing that `tox` suite does is parse the `setup.py` file, even before installing any of the actual env dependencies listed in `tox.ini`. Therefore, adding the `setuptools` constraint to `deps` in tox.ini, or to `install_requires`, or to `setup_requires`, seemingly does nothing, because `tox` will immediately launch with a setuptools version which cannot handle the `setup.py`, and immediately fail out. Since tox fails before parsing the actual package deps and the nested configuration doesn't work due to the previous issue, the solution seems to be to simply upgrade `setuptools` on the build machine for py35 before launching anything. That way, `tox` sees that its requirements are already satisfied, and does not use the shim.

I ended up adding a pyproject.toml which notates this constraint, so in practice, the generated package will have the proper dependency on `setuptools` already configured. This means that a user doing an actual install shouldn't run into this, because the generated package will already depend on the correct version of `setuptools`.

alexweav:
FYI @DavidCurtiss

setup.py
spanglerco: So now that we have a pyproject.toml, do we still need this too?

DavidCurtiss:
👍 Looks good.

.github/workflows/python-package.yml
```diff
@@ -22,7 +22,7 @@ jobs:
         python-version: ${{ matrix.python-version }}
     - name: Install dependencies
       run: |
-        python -m pip install --upgrade pip
+        python -m pip install --upgrade pip setuptools wheel
```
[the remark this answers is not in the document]
alexweav: I'm unsure, but it doesn't hurt to split the command. Pip doesn't officially guarantee an install order if there aren't dependencies between the requested packages.

setup.py
alexweav: Don't believe so. I'll double check that the dependency is still there.

PR #13, opened 2020-09-22
[setup.py] Prevent tests from being included in .whl

alexweav:

Manually open all prior versions of the .whl and notice that it contains a `tests` module. These files are also installed when `pip install`ing this package. We don't need to distribute tests, and they bloat the package unnecessarily.

This change updates the package exclusion rules to properly leave out everything under the `tests` directory.

In previous versions, there was an exclusion rule for `tests`, but this only instructs setup.py to exclude everything directly in the `tests` module, but does not exclude submodules. To exclude a module and its submodules, both `mymodule` and `mymodule.*` should be excluded - see [these docs](https://setuptools.readthedocs.io/en/latest/setuptools.html#using-find-packages).

Examples were already not included because they do not contain an `__init__.py`. However for future-proofing, I explicitly added examples to the exclusion rules as well, in case an init is ever added.

Built the .whl. Verified that `tests` was no longer present.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -17,7 +17,7 @@ class PyTest(TestCommand):
 pypi_name = "nisystemlink-clients"
 
 packages = find_namespace_packages(include=["systemlink.*"]) + find_packages(
-    exclude=["systemlink", "examples", "examples.*", "tests", "tests.*"]
+    exclude=["systemlink", "systemlink.*", "examples", "examples.*", "tests", "tests.*"]
 )
 
 
--- a/setup.py
+++ b/setup.py
@@ -17,7 +17,7 @@ class PyTest(TestCommand):
 pypi_name = "nisystemlink-clients"
 
 packages = find_namespace_packages(include=["systemlink.*"]) + find_packages(
-    exclude=["systemlink", "examples", "tests"]
+    exclude=["systemlink", "examples", "examples.*", "tests", "tests.*"]
 )
 
 
--- a/setup.py
+++ b/setup.py
@@ -16,9 +16,7 @@ class PyTest(TestCommand):
 
 pypi_name = "nisystemlink-clients"
 
-packages = find_namespace_packages(include=["systemlink.*"]) + find_packages(
-    exclude=["systemlink", "systemlink.*", "examples", "examples.*", "tests", "tests.*"]
-)
+packages = find_namespace_packages(include=["systemlink.*"])
 
 
 def _get_version(name):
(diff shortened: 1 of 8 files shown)


alexweav:
FYI @DavidCurtiss

alexweav:
Not sure why Black is now failing on a bunch of unrelated files. Perhaps they updated and we don't have a pinned version?

DavidCurtiss:
> Not sure why Black is now failing on a bunch of unrelated files. Perhaps they updated and we don't have a pinned version?

Yep, that's exactly it. I just reproduced by first running "pip install --upgrade black". (`black --check` passed before upgrading, and failed afterwards. My prior installed version was 19.10b0.)

In all cases, there was a comma at the end of a single-line list (example below), which black now takes to mean that you want it split across lines.
```python
return self._handle_read(path, response3, http_response,)
```

You should remove the trailing comma in all those cases, and then both versions of black will be happy with it.

DavidCurtiss (approved):

spanglerco:
Should we be pinning versions to avoid issues like this? It seems like upgrading dependencies should be an intentional process with a dedicated PR so it can resolve things like this.

setup.py
```diff
@@ -17,7 +17,7 @@ def run_tests(self):
 pypi_name = "nisystemlink-clients"
 
 packages = find_namespace_packages(include=["systemlink.*"]) + find_packages(
-    exclude=["systemlink", "examples", "tests"]
+    exclude=["systemlink", "systemlink.*", "examples", "examples.*", "tests", "tests.*"]
```
spanglerco, line 20: I might be misunderstanding what this is doing, but what is there besides these? Is the `find_packages` call actually returning anything that `find_namespace_packages` isn't?

alexweav:
@spanglerco The Python working group has discussed the use of `poetry`, and will almost certainly end up recommending it, there is a lot of support for it across NI. Once it is officially recommended as the toolchain, we should be looking to migrate projects to be compliant.

setup.py
alexweav: The intent was to have this statement be a "catch-all" for all normal packages, i.e. the ones which don't have special behavior due to namespace package rules. It semantically matches the previous implementation where all valid packages in the repo are included except for a select set of directories.

Indeed, it is redundant in the current package layout, though. Might as well remove it.

spanglerco (approved):

DavidCurtiss (approved):

PR #15, opened 2020-12-15
send tag writes on buffered tag exit

### LINKED ISSUE #14: Tag writes are not automatically sent on exit of the buffered tag writer
The basic [read_write_one_tag example](https://github.com/ni/nisystemlink-clients-python/blob/master/examples/tag/read_write_one_tag.py) doesn't work. The "Note: Exiting the 'with' block automatically calls writer.send_buffered_writes()" appears to be incorrect: exiting the with block is not sending the writes. Instead, you have to manually call `writer.send_buffered_writes()` (or use a buffer size of 1).

adriandeac:

This makes the buffered tag writer send any buffered writes on exit.

Resolves #14 

Ran all the tests that didn't need a local instance of SystemLink Server or the SystemLink Cloud setup.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -51,7 +51,7 @@ setup(
     packages=packages,
     install_requires=[
         'aenum;python_version<"3.6"',
-        "events==0.3",
+        "events",
         'httpx;python_version>="3.6"',
         'requests;python_version<"3.6"',
         "typing-extensions",
--- a/setup.py
+++ b/setup.py
@@ -51,7 +51,7 @@ setup(
     packages=packages,
     install_requires=[
         'aenum;python_version<"3.6"',
-        "events",
+        "events==0.3",
         'httpx;python_version>="3.6"',
         'requests;python_version<"3.6"',
         "typing-extensions",
--- a/systemlink/clients/tag/_buffered_tag_writer.py
+++ b/systemlink/clients/tag/_buffered_tag_writer.py
@@ -198,7 +198,6 @@ class BufferedTagWriter(tbase.ITagWriter):
             return False
 
         self.send_buffered_writes()
-        self._timer_generation += 1
         self._closed = True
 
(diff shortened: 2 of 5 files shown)


setup.py
```diff
@@ -51,7 +51,7 @@ def _read_contents(file_to_read):
     packages=packages,
     install_requires=[
         'aenum;python_version<"3.6"',
-        "events",
+        "events==0.3",
```
adriandeac, line 54: Note: I think [this change](https://github.com/pyeve/events/commit/e1895e27a6f87417fb87633d444d5a2562994b16) broke a few tests that were asserting on instances of `HttpTagSubscription` so locking down the version to `0.3` for now.

systemlink/clients/tag/_buffered_tag_writer.py
```diff
@@ -197,6 +197,7 @@ def __exit__(
         if self._closed:
             return False
 
+        self.send_buffered_writes()
         self._timer_generation += 1
```
spanglerco, line 201: We should be able to remove this (and the one in `__aexit__` below) since `send_buffered_writes` will do the same thing.
```suggestion
```

spanglerco (approved):

setup.py
DavidCurtiss: Hmm. That's a weird behavior. I think a better fix may be to define our own `__bool__` method that returns `True`, and/or update the tests to "assert uut is not None" instead of just "assert uut".

DavidCurtiss (approved):

systemlink/clients/tag/_buffered_tag_writer.py
adriandeac: Done.

spanglerco: Thanks

PR #18, opened 2022-11-07
Create new "dataframe" subpackage

mure:

Creates a subpackage "dataframe" that provides a new client for the DataFrame service.

We need a place to organize the dataframe modules.

n/a

### DIFF
--- a/systemlink/clients/dataframe/__init__.py
+++ b/systemlink/clients/dataframe/__init__.py
@@ -1,0 +1,3 @@
+from ._data_frame_client import DataFrameClient
+
+# flake8: noqa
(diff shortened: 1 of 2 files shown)


mure:
I created the class `DataFrameClient` which strays from the convention of the tag client that has a `TagManager` class instead. Does anyone have strong opinions that we should keep the "manager" convention? It would be something like `DataTableManager` instead.

cameronwaterman:
> I created the class `DataFrameClient` which strays from the convention of the tag client that has a `TagManager` class instead. Does anyone have strong opinions that we should keep the "manager" convention? It would be something like `DataTableManager` instead.

I agree with moving away from using "manager" in the client name. It really has no significance to the client itself.

cameronwaterman (approved):

spanglerco (approved):
Are we still planning to rename the package to nisystemlink?

mure:
> Are we still planning to rename the package to nisystemlink?

Yes, that'll be in another PR.

PR #20, opened 2022-11-09
Add uplink dependency and new "internal" package with BaseClient

mure:

- Adds dependencies on `uplink` and `pydantic`, which will be used to implement the new DFS client.
- Adds a new private package to "core" to contain our uplink helpers and wrappers
- Creates `BaseClient` class that extends from [uplink.Consumer](https://uplink.readthedocs.io/en/stable/user/quickstart.html)
- Adds `DataFrameClient` class with a rudimentary implementation of the API info route for integration testing purposes 

Needed for DFS client

This will be tested by integration tests later on.


systemlink/clients/dataframe/_data_frame_client.py
```diff
@@ -1,5 +1,16 @@
+# TODO: Wrap Uplink decorators to add typing information
```
spanglerco, line 1: FYI: It's also possible to create a pyi stub like [this one](https://github.com/ni/nisystemlink-clients-python/blob/master/systemlink/stubs/events.pyi) that adds typing information to a library without modifying that library. It may not be worth it though... But there's [a command you can run](https://mypy.readthedocs.io/en/stable/stubs.html) to generate the initial stub file for you to fill in.

spanglerco (approved):

systemlink/clients/dataframe/_data_frame_client.py
mure: There's a good chance I'll end up wrapping their decorators anyway to simplify things for our clients. But the stubs could be helpful. I'll look into it.

cameronwaterman (approved):

nisystemlink/clients/dataframe/_data_frame_client.py
```diff
@@ -3,11 +3,27 @@
...
+    def __init__(self, configuration: Optional[core.HttpConfiguration] = None):
+        """Initialize an instance.
+
+        Args:
+            configuration: Defines the web server to connect to and information about
+                how to connect. If not provided, an instance of
+                :class:`core.JupyterHttpConfiguration` is used.
+        """
+        if configuration is None:
+            configuration = core.JupyterHttpConfiguration()
```
spanglerco, line 23: Should this go in the base client so we have consistent behavior? I imagine we'll also want a fallback where if the environment variables are missing, we look for files (i.e. add Jupyter support to HttpConfigurationManager).

PR #22, opened 2022-11-10
Rename package to nisystemlink

cameronwaterman:

Renames the package to "nisystemlink" since the existing name "systemlink" conflicts with the SystemLink SDK which would prevent it from being used in Jupyter.

Provides a unique package name so it can be used in Jupyter.

- Ran tests
- Ran linter
- Built package and inspected ouput


pyproject.toml
```diff
@@ -43,9 +43,10 @@ poethepoet               = "^0.16.4"
 [tool.poe.tasks]
 test       = "pytest tests -m \"(not slow) and (not cloud)\""
 doctest    = "pytest --doctest-modules docs"
-format     = "black --exclude \".*\\.pyi\" --check systemlink examples tests"
-lint       = "flake8 systemlink examples tests"
-types      = "mypy --config-file mypy.ini systemlink examples/tag tests"
+check      = "black --exclude \".*\\.pyi\" --check nisystemlink examples tests"
```
cameronwaterman, line 46: @mure - I changed "format" to "check" and updated the workflow to call this one instead, and then added a new "format" task that actually performs the format.

mure: Makes sense - thanks

mure (approved):

nisystemlink/clients/core/_internal/_winpaths.pyi
```diff
@@ -4,11 +4,9 @@ import ctypes
 from ctypes import wintypes
 from uuid import UUID
 
-
```
spanglerco, line 7: I'm curious about tehse edits come from, and for stubs/events.pyi as well. Aren't there usually two blank lines between classes? Also, it looks as though the black command is intended to exclude pyi files.

spanglerco (approved):

nisystemlink/clients/core/_internal/_winpaths.pyi
cameronwaterman: These changes did come from black, (running `poe format`) - so it looks like there might be an issue with the regex that is supposed to exclude pyi files

cameronwaterman: Looks like there was a Windows-specific issue when specifying the `--exclude` regex inline in the poe task. Moving the exclude config out to a black-specific section in the pyproject.toml file seems to have fixed it.

PR #23, opened 2022-11-11
Handle HTTP errors and add model for api_info method

mure:

- Adds a response handler that catches HTTP errors and raises `ApiException` instances
- Base model for serializing/deserializing HTTP requeset/response bodies using Pydantic
- Model for `api_info` method response

Tested manually locally. Example:

```python
from devtools import debug
from nisystemlink.clients.core import HttpConfiguration
from nisystemlink.clients.dataframe import DataFrameClient

client = DataFrameClient("https://api-stratus-test.ion.ni.com/", "foobar"))

debug(client.api_info())
```
Output:
```python
scratch.py:12 <module>
    client.api_info(): ApiInfo(
        operations=OperationsV1(
            create_tables=Operation(
                available=True,
                version=1,
            ),
            delete_tables=Operation(
                available=True,
                version=1,
            ),
            modify_metadata=Operation(
                available=True,
                version=1,
            ),
            list_tables=Operation(
                available=True,
                version=1,
            ),
            read_data=Operation(
                available=True,
                version=1,
            ),
            write_data=Operation(
                available=True,
                version=1,
            ),
        ),
    ) (ApiInfo)
```


nisystemlink/clients/core/_uplink/_base_client.py
```diff
@@ -1,5 +1,29 @@
...
+        if response.status_code == 204:
+            return None
+        return response
+    try:
+        content = response.json()
+        if isinstance(content, dict) and "error" in content:
+            error = core.ApiError.from_json_dict(content["error"])
+            raise core.ApiException(error=error, http_status_code=response.status_code)
+        else:
+            cast(Callable[[], NoReturn], response.raise_for_status)()
```
mure, line 24: mypy was complaining that this method was missing returns because `raise_for_status` wasn't typed as `NoReturn`, so I had to add the cast

nisystemlink/clients/core/_uplink/_json_model.py
```diff
@@ -0,0 +1,16 @@
...
+    return next(parts) + "".join(i.title() for i in parts)
+
+
+class JsonModel(BaseModel):
+    """Base class for models that are serialized to and from JSON."""
+
+    class Config:
+        alias_generator = _camelcase
+        allow_population_by_field_name = True
+        extra = Extra.forbid
```
mure, line 16: The `Extra.forbid` config makes it so Pydantic will error if the model receives an unknown field when initializing. I think this is valuable, because it will make our integration tests fail when we add new fields to any resources. It also will work nicely with the `ApiInfo` model below, causing errors if we add new operations without updating the client.

nisystemlink/clients/dataframe/_data_frame_client.py
```diff
@@ -25,8 +27,7 @@ def __init__(self, configuration: Optional[core.HttpConfiguration] = None):
         super().__init__(configuration)
 
     @returns.json()
-    @get("nidataframe")
-    # TODO: Create model class for return type
-    def api_info(self) -> dict:
-        """Returns information about API versions and available operations."""
+    @get("nidataframe/v1")
```
mure, line 30: I changed this method to be version specific. I think it makes more sense to pin the client implementation to a single version of the API. It's also kind of weird to expose the concept of versions when its not relevant to the usage of the client methods (unless we want to explicitly have, for example, `list_tables_v1()`, which I'm not in favor of).

spanglerco: Would it make sense to have a constant or something for the "nidataframe/v1" part since every URL will have it?

nisystemlink/clients/dataframe/models/_api_info.py
```diff
@@ -0,0 +1,19 @@
+from nisystemlink.clients.core._uplink._json_model import JsonModel
+
+
+class Operation(JsonModel):
```
spanglerco, line 4: Should these classes have docstrings?

spanglerco, line 4: Do we need to add these classes to [\_\_init\_\_.py](https://github.com/ni/nisystemlink-clients-python/blob/0b2bf8e8df6290cdba62a2538ebd0259718edd55/nisystemlink/clients/dataframe/__init__.py)?

spanglerco (changes requested):

nisystemlink/clients/core/_uplink/_json_model.py
mure: I wonder if we can monkey patch the config to turn it on during integration tests. That would be the best of both worlds.

nisystemlink/clients/dataframe/models/_api_info.py
mure: Currently, they'll be available under the `models` namespace, e.g:

`from nisystemlink.clients.dataframe.models import ApiInfo`

Do we want to make them available from the `dataframe` namespace?

mure: Yes, thanks. flake8 isn't complaining because all of these modules are technically private 😕

nisystemlink/clients/dataframe/_data_frame_client.py
mure: Done

cameronwaterman (approved):

nisystemlink/clients/dataframe/models/_api_info.py
spanglerco: Oh, I missed the `from ._api_info import *` in the model's \_\_init\_\_.py. Yeah, that should be fine.

spanglerco: Nice. Do we have the ability to specify a docstring for the attributes too? I'm not sure what sphinx looks for.

mure: > I'm not sure what sphinx looks for.
I'm testing that right now actually. I just enabled the docs for the dataframe package to see how attributes work.

mure: I improved the generated docs for the DFS client and pulled in [autodoc-pydantic](https://autodoc-pydantic.readthedocs.io/en/stable/index.html) for the models. There's still more work we can do to clean up the docs, but I'll stop here for this PR.

spanglerco (approved):

nisystemlink/clients/core/_uplink/_json_model.py
spanglerco: That could be interesting, especially if it were an option to toggle it on or off. I could see cases where you make a change to the service that adds a new field and then someone wants to make an unrelated change to the client code and tests start failing.

nisystemlink/clients/core/_uplink/_base_client.py
mure: I refactored this to be more like how the tag API handles things.

PR #24, opened 2022-11-11
Data Frame integration test infrastructure

cameronwaterman:

Adds integration test infrastructure for testing the Data Frame client.

We need to be able to integration test the Data Frame client to ensure functionality against a production service.

Ran existing and new tests locally


.github/workflows/python-package.yml
```diff
@@ -21,7 +21,9 @@ jobs:
           python-version: ${{ matrix.python-version }}
           cache: "poetry"
       - run: poetry install
-      - run: poetry run pytest --cloud-api-key ${{ secrets.CLOUD_API_KEY }}
+      - run: poetry run pytest
+      - run: poetry run pytest -m cloud --cloud-api-key "${{ secrets.CLOUD_API_KEY }}"
+      - run: poetry run pytest -m enterprise --enterprise-api-key "${{ secrets.ENTERPRISE_API_KEY }}"
```
spanglerco, line 26: I think we should also specify the URL as a parameter instead of hard-coding dev in the code, so there's only one place to change it and so developers can test against different clusters (e.g. a customer could even run the tests against their own version). In particular, I wonder if we should use test instead of dev.

pyproject.toml
```diff
@@ -56,6 +56,7 @@ markers = [
     "unit: mark a test as a unit test",
     "integration: mark a test as an integration test",
     "cloud: mark a test as a cloud integration test",
+    "enterprise: mark a test as a enterprise integration test",
```
spanglerco, line 59:
```suggestion
    "enterprise: mark a test as an enterprise integration test",
```

.github/workflows/python-package.yml
spanglerco, line 26: Any specific reason not to run all of the tests in one command instead of splitting it into 3? My understanding is if you pass both cloud and enterprise arguments, it'll be able to run all of the tests.

spanglerco (changes requested):

nisystemlink/clients/core/_enterprise_http_configuration.py
```diff
@@ -0,0 +1,22 @@
+# -*- coding: utf-8 -*-
+
+"""Implementation of EnterpriseHttpConfiguration."""
+
+from nisystemlink.clients import core
+
+
+class EnterpriseHttpConfiguration(core.HttpConfiguration):
```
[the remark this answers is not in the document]
cameronwaterman: We use the term "enterprise" as a marker for the tests as well, what should we do about that case?

.github/workflows/python-package.yml
cameronwaterman: Should we do the same for the cloud tests?

Also, I chose dev because that is what we are using for our E2E tests.

nisystemlink/clients/core/_enterprise_http_configuration.py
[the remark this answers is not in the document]
spanglerco: I think test markers and parameters are fine since they don't appear in the API, just the test code. We can always rename them later if a different name makes more sense.

.github/workflows/python-package.yml
spanglerco: I think SystemLink Cloud is different, because it only exists at systemlinkcloud.io, though there are technically multiple tiers as well.

The E2E tests are testing the product itself, which makes sense to do at each tier, while these tests are for the clients. In the future, we could add an E2E test that pulls and runs these tests against each tier as well, to detect API breaks.

.github/workflows/python-package.yml
```diff
@@ -21,7 +21,7 @@ jobs:
           python-version: ${{ matrix.python-version }}
           cache: "poetry"
       - run: poetry install
-      - run: poetry run pytest --cloud-api-key ${{ secrets.CLOUD_API_KEY }}
+      - run: poetry run pytest --cloud-api-key "${{ secrets.CLOUD_API_KEY }}" --enterprise-uri "https://test-api.lifecyclesolutions.ni.com" --enterprise-api-key "${{ secrets.ENTERPRISE_API_KEY }}"
```
mure, line 24: Nit: It would be nice if this was split across multiple lines

mure (approved):

spanglerco (approved):

PR #26, opened 2022-11-14
Disallow extra Pydantic fields in tests

mure:

Adds a session-level pytest fixture that sets the ["extra" config in Pydantic](https://pydantic-docs.helpmanual.io/usage/model_config/#options) models to "forbid".

Using this fixture will cause our tests to fail when new fields are added to DFS resources. This will force us to keep the client in sync with the service.

Ran tests locally and verified that it causes failures as expected.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -87,5 +87,5 @@ def enterprise_config(pytestconfig):
 
 @pytest.fixture(scope="session", autouse=True)
 def pydantic_forbid_extra_fields():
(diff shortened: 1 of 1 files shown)


spanglerco (approved):
I wonder if this may become awkward in the future. I'm imagining updating DFS with a new response property, but it hasn't been released yet, so we may not want to add it to the client yet (or we just haven't gotten around to it yet). Then someone wants to make a change to the tag client, but the PR build fails because of the update to DFS server-side.

PR #28, opened 2022-11-15
feat: Implement create/get/delete/list table metadata methods

mure:

- Implements a set of table metadata routes:
  - create
  - get
  - delete
  - list

- Adds a custom serializer for our Pydantic models. This is so we can set certain options like `by_alias`.
- Switches attribute comments to use docstrings, because some of them wouldn't fit on a single line.

We need this functionality.
(This part of the PR template is awkward. I'm never sure what to put other than "because we need it")

Added our first functional test 🎉 Feel free to suggest other test cases. I think we'll want at least one that tests an expected error.


mypy.ini
```diff
@@ -13,7 +13,6 @@ strict_equality=True
 
 [mypy-tests.*]
 disallow_untyped_calls=True
-disallow_incomplete_defs=True
```
mure, line 16: This was forcing me to add `-> None` to every pytest method, which seemed silly.

docs/conf.py
```diff
@@ -57,3 +57,4 @@
 
 # Configuration for https://autodoc-pydantic.readthedocs.io
 autodoc_pydantic_model_show_config_summary = False
+autodoc_pydantic_field_show_alias = False
```
mure, line 60: We use Pydantic aliases purely so we can deserialize camelCase to snake_case and vice versa, so there's no reason to display them in the docs.

cameronwaterman (changes requested):
Please change your title to adhere to the [Angular style](https://github.com/angular/angular.js/blob/master/DEVELOPERS.md#commits)

i.e. "feat: Implement create/get/delete table metadata methods"

It looks like there is also a typo in your existing title 😆

cameronwaterman (approved):

nisystemlink/clients/dataframe/models/_column.py
```diff
@@ -0,0 +1,22 @@
...
+class Column(JsonModel):
+    """Defines a single column in a table."""
+
+    name: str
+    """The column name, which must be unique across all columns in the table."""
+
+    data_type: DataType
+    """The data type of the column."""
+
+    column_type: Optional[ColumnType] = ColumnType.Normal
```
spanglerco, line 18: If you set `column_type` to None, would the serializer pass `"columnType": null` in the JSON, or leave it out of the JSON? The service will reject null, but will default to normal if you leave it out.

I ask because we may want to remove the `Optional` part if None would map to null in the JSON. Or if there's a way to annotate the model such that the serializer will leave out the key if it's None, then we could set it to none by default and let the server choose the default value instead of the client. But I don't know best practice on letting the client choose the default vs. the server.

spanglerco (commented):
Some test ideas:

- Validate we deserialize timestamps correctly by asserting the created_at on the table metadata. We would probably need some kind of threshold for allowed values given that the time on the server may not match the time on the client.
- Retrieve metadata for a valid table ID that doesn't exist (like `000000000000000000000000`) and verify we raise the correct exception that includes the error details like the code and args.
- Similarly, we could have an error case for the create and delete APIs.

nisystemlink/clients/dataframe/models/_column.py
mure: This is a good catch. It's serializing it to `null` and throwing an error. I removed the `Optional` annotation. 

In general, letting the server choose defaults is probably the better approach, but I wanted the default here so that the `Column` class could be reused for both responses and requests. Also, it's very unlikely the default will ever change from Normal.

mure:
@spanglerco @cameronwaterman I was working off this branch and since the PR is still up, I went ahead and added the list_tables method as well. I'll reset the reviews so y'all can take a look.

tests/integration/dataframe/test_dataframe.py
```diff
@@ -1,49 +1,76 @@
...
+        id = client.create_table(table)
+        print(f"Created table {id}")
+        tables.append(id)
+        return id
+
+    yield _create_table
+
+    for id in tables:
+        client.delete_table(id)
+        print(f"Deleted table {id}")
```
spanglerco, line 32: We may want to leave a note to change this to use the multi-delete route when we have it.

spanglerco (changes requested):

nisystemlink/clients/dataframe/models/_order_by.py
```diff
@@ -0,0 +1,14 @@
+from typing import Literal
+
+
+OrderBy = Literal[
```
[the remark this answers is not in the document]
mure: I'm working around this issue: https://github.com/prkumar/uplink/issues/281. I added a comment to make this clear.

nisystemlink/clients/dataframe/_data_frame_client.py
```diff
@@ -33,6 +33,41 @@ def api_info(self) -> models.ApiInfo:
...
+            Query("orderBy"),
+            Query("orderByDescending"),
+            Query("continuationToken"),
+            Query("workspace"),
+        ),
+    )
+    def list_tables(
+        self,
+        take: Optional[int] = None,
+        id: Optional[List[str]] = None,
```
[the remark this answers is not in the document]
mure: Looks like it's serializing it the first way `?id=1&id=2` (and the service handles it correctly). Do we need to fix it to be comma separated?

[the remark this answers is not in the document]
cameronwaterman: @spanglerco - This is the last remaining open issues on this PR and we're wanting to get this submitted to test out the semantic release stuff. I spoke with Carson offline and we agreed that since this works in its existing state we can go ahead and submit this change with this behavior, especially since we're still publishing prerelease packages. I do think we should consider making this change though.

@mure - Please create a follow-up task under your user story to discuss whether or not we should address this in a separate PR.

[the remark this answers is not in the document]
spanglerco: If there's an easy way to get uplink to send it as a comma-separated list, I would prefer that. But if it's a lot of work, we probably don't need to worry about it. We may want to add tests to the DFS to ensure we don't break this way of doing it. We're getting it for free from ASP.NET, so if we were to switch frameworks in the future, it may break without us thinking about it.

PR #33, opened 2022-11-28
feat: Method to update table metadata

mure:

- Adds `update_table_metdata`method to patch table metdata.
The behavior of setting `null` to remove properties/fields relies on Pydantic's `exclude_unset` config: https://pydantic-docs.helpmanual.io/usage/exporting_models/

- Added automated test

### Open Issues

Does anyone have input on the naming for the patch methods? I wasn't sure what to go with.
- `modify_table_metadata`
- `modify_table(s)`
- `patch_table(s)`

Keep in mind that we will need a name for the multi-patch method as well.

### DIFF
--- a/nisystemlink/clients/dataframe/_data_frame_client.py
+++ b/nisystemlink/clients/dataframe/_data_frame_client.py
@@ -109,7 +109,7 @@ class DataFrameClient(BaseClient):
 
     @json
     @patch(_BASE_PATH + "/tables/{id}", args=(Path, Body))
-    def update_table_metadata(self, id: str, update: models.ModifyTableRequest) -> None:
+    def modify_table(self, id: str, update: models.ModifyTableRequest) -> None:
         """Modify properties of a table or its columns.
 
         Args:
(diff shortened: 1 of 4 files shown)


spanglerco:
> Does anyone have input on the naming for the patch methods?

In the DFS code, we call the patch route `ModifyTable` and the multi-edit route `ModifyTables`. So I think `modify_table` and `modify_tables` could work here too.

nisystemlink/clients/dataframe/_data_frame_client.py
```diff
@@ -107,6 +107,17 @@ def get_table_metadata(self, id: str) -> models.TableMetadata:
         """
         ...
 
+    @json
+    @patch(_BASE_PATH + "/tables/{id}", args=(Path, Body))
+    def update_table_metadata(self, id: str, update: models.ModifyTableRequest) -> None:
```
spanglerco, line 112: Regardless of what we go with for the method name, I think we should be consistent between `update_table_metadata` and `ModifyTableRequest` (update vs. modify). In the DFS OpenAPI document, we consistently use "modify".

spanglerco (approved):

nisystemlink/clients/dataframe/_data_frame_client.py
mure: I'm going with `modify_table(s)`.

PR #35, opened 2022-12-02
feat: Add modify_tables method

mure:

- Adds method for patching multiple tables in one call
- `BaseClient` takes in base path so clients don't have to manually build the URI
- Extracted simple table model to be reduce test verbosity

- Added two automated tests for `modify_tables`


nisystemlink/clients/dataframe/models/_table_metadata_modification_base.py
```diff
@@ -0,0 +1,29 @@
...
+    state."""
+
+    name: Optional[str] = None
+    """The new name of the table. Setting to ``None`` will reset the name to the table's ID."""
+
+    workspace: Optional[str] = None
+    """The new workspace for the table. Setting to ``None`` will reset to the
+    default workspace. Changing the workspace requires permission to delete the
+    table in its current workspace and permission to create the table in its new
+    workspace."""
```
spanglerco, line 24: The "setting to None will reset" behavior is only applicable to the PATCH route. The multi-edit POST route considers null to mean unchanged as if it weren't present in the request.

nisystemlink/clients/dataframe/models/_modify_tables_partial_success.py
```diff
@@ -0,0 +1,19 @@
+from typing import List
+
+from nisystemlink.clients.core import ApiError
+from nisystemlink.clients.core._uplink._json_model import JsonModel
+
+from ._modify_tables_request import TableMetdataModification
+
+
+class ModifyTablesPartialSuccess(JsonModel):
+    """The result of deleting multiple tables when one or more tables could not be deleted."""
```
spanglerco, line 10:
```suggestion
    """The result of modifying multiple tables when one or more tables could not be modified."""
```

tests/integration/dataframe/test_dataframe.py
```diff
@@ -20,7 +30,7 @@ def create_table(client: DataFrameClient):
     tables = []
 
     def _create_table(table: models.CreateTableRequest) -> str:
```
spanglerco, line 32: Given the `or basic_table_model` below, did you intend to make `table` optional?

```suggestion
    def _create_table(table: Optional[models.CreateTableRequest] = None) -> str:
```

(with an import being required)

spanglerco (changes requested):

nisystemlink/clients/dataframe/models/_table_metadata_modification_base.py
mure: Yeah I realized that as I was writing the tests, but I didn't think about the docstrings. Maybe this base class isn't worth it.

mure: Why is the behavior different anyway?

spanglerco (approved):

PR #36, opened 2022-12-05
feat: Read and append table data

mure:

- `append_table_data` and `get_table_data` implemented

- Added automated test that writes data and then reads back a **mocked** response. The mock is a temporary solution until the service supports manual flushing of data.
- Manually tested reading data

### Questions for reviewers

Any opinions on the method names? I'm open to alternatives. (`write_table_data`, `read_table_data`)

Are there other test cases we should cover?

### DIFF
--- a/nisystemlink/clients/dataframe/models/_data_frame.py
+++ b/nisystemlink/clients/dataframe/models/_data_frame.py
@@ -43,11 +43,10 @@ class DataFrame(JsonModel):
     https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.to_json.html.
 
     When providing a DataFrame for appending rows, any of the table's columns
-    not specified columns will receive a value of ``None``. If any such columns
-    aren't nullable, an error will be returned. If the entire columns property
-    is left out, each row is assumed to contain all columns in the order
-    specified when the table was created.
-
+    not specified will receive a value of ``None``. If any such columns aren't
+    nullable, an error will be returned. If the entire columns property is left
+    out, each row is assumed to contain all columns in the order specified when
+    the table was created.
     """
 
     columns: Optional[List[str]] = None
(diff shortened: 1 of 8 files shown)


nisystemlink/clients/dataframe/models/_data_frame.py
```diff
@@ -0,0 +1,58 @@
...
+
+    The format is the same as a serialized Pandas DataFrame with orient="split"
+    and index=False. See
+    https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.to_json.html.
+
+    When providing a DataFrame for appending rows, any of the table's columns
+    not specified columns will receive a value of ``None``. If any such columns
+    aren't nullable, an error will be returned. If the entire columns property
+    is left out, each row is assumed to contain all columns in the order
+    specified when the table was created.
```
spanglerco, line 49: Looks like an extra "columns" in there:

```suggestion
    When providing a DataFrame for appending rows, any of the table's columns
    not specified will receive a value of ``None``. If any such columns aren't
    nullable, an error will be returned. If the entire columns property is left
    out, each row is assumed to contain all columns in the order specified when
    the table was created.
```

tests/integration/dataframe/test_dataframe.py
```diff
@@ -276,3 +277,60 @@ def test__modify_tables__returns_partial_success(self, client: DataFrameClient):
...
+            )
+        )
+
+        frame = models.DataFrame(
+            columns=["index", "value", "ignore_me"],
+            data=[["1", "3.3", "True"], ["2", None, "False"], ["3", "1.1", "True"]],
+        )
+
+        client.append_table_data(
+            id, models.AppendTableDataRequest(frame=frame, end_of_data=True)
```
spanglerco, line 310: Should we add an error case, either passing invalid data or appending to a table after marking it end of data?

spanglerco (approved):

tests/integration/dataframe/test_dataframe.py
mure: Added a test that tries to write to a column that doesn't exist

cameronwaterman (approved):

PR #38, opened 2022-12-07
Documentation fixes

mure:

- Properly documents methods of `DataFrameClient`
The uplink decorators turns the functions into objects and Sphinx isn't auto documenting them meaningfully, even though the special methods like `__doc__` are preserved.
image
After some trial and error, I discovered that manually specifying the methods with `automethod::` gets Sphinx to treat the objects like functions and generate the correct docs.

- Reduces the clutter on the index page by reducing the TOC depth

It makes the docs better!

- See the built docs from this PR

### DIFF
--- a/docs/api_reference/dataframe.rst
+++ b/docs/api_reference/dataframe.rst
@@ -6,6 +6,8 @@ nisystemlink.clients.dataframe
 .. autoclass:: nisystemlink.clients.dataframe.DataFrameClient
    :members:
 
+   .. automethod:: delete_table(id)
+
 .. automodule:: nisystemlink.clients.dataframe.models
    :members:
    :imported-members:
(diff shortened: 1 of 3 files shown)


docs/api_reference/dataframe.rst
```diff
@@ -3,9 +3,23 @@
 nisystemlink.clients.dataframe
 ======================
 
-.. automodule:: nisystemlink.clients.dataframe
-   :members:
-   :imported-members:
+.. autoclass:: nisystemlink.clients.dataframe.DataFrameClient
+   :exclude-members: __init__
```
mure, line 7: The ordering was getting messed up with this fix 😕
In `conf.py`, we set a sphinx autodoc option to always document the special member `__init__`. But it was last in the list after the client methods, which isn't expected. The way I fixed it was by manually excluding `__init__` from the autoclass and then specifying it as an automethod.

cameronwaterman (approved):

spanglerco (approved):
This reminds me that we may want a work item to add a getting started section for DFS, along with an example.

cameronwaterman:
Whoops - forgot the "docs:" tag in the title. Shouldn't matter as it doesn't produce a new release for docs I don't think.

PR #40, opened 2022-12-12
docs: Add examples and update getting started

cameronwaterman:

- Add `DataFrameClient` usage examples
- Update "Getting Started" docs with `DataFrameClient` overview and examples

Good docs = happy customers

- Verified examples work as expected in Jupyter notebook
- Ran `poe` tasks locally
- Verified docs look correct and links work

### DIFF
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -75,3 +75,45 @@ Subscribe to tag changes
 .. literalinclude:: ../examples/tag/subscribe_to_tag_changes.py
    :language: python
    :linenos:
+
+Data Frame API
+-------
+
+Overview
+~~~~~~~~
+
+The :class:`.DataFrameClient` class is the primary entry point of the Data Frame API.
+
+When constructing a :class:`.DataFrameClient`, you can pass an
+:class:`.HttpConfiguration` (like one retrieved from the
+:class:`.HttpConfigurationManager`), or let :class:`.DataFrameClient` use the
+default connection. The default connection depends on your environment.
+
+With a :class:`.DataFrameClient` object, you can:
+
+* Retrieve information about available API operations
+
+* Query, create, modify, and delete Data Frame Tables and associated metadata
+
+* Query, query decimated, and append Data Frame Table data
+
+Examples
+~~~~~~~~
+
+Create and write data to a table
+
+.. literalinclude:: ../examples/dataframe/create_write_data.py
+   :language: python
+   :linenos:
+
+Query and read data from a table
+
+.. literalinclude:: ../examples/dataframe/query_read_data.py
+   :language: python
+   :linenos:
+
+Delete a table
+
+.. literalinclude:: ../examples/dataframe/delete_table.py
+   :language: python
+   :linenos:
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -91,11 +91,11 @@ default connection. The default connection depends on your environment.
 
 With a :class:`.DataFrameClient` object, you can:
 
-* Retrieve information about available API operations
+* Create and delete Data Frame Tables.
 
-* Query, create, modify, and delete Data Frame Tables and associated metadata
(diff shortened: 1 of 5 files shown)


docs/getting_started.rst
```diff
@@ -75,3 +75,45 @@ Subscribe to tag changes
...
+
+When constructing a :class:`.DataFrameClient`, you can pass an
+:class:`.HttpConfiguration` (like one retrieved from the
+:class:`.HttpConfigurationManager`), or let :class:`.DataFrameClient` use the
+default connection. The default connection depends on your environment.
+
+With a :class:`.DataFrameClient` object, you can:
+
+* Retrieve information about available API operations
+
```
spanglerco, line 95: I'd be fine with leaving this one out. It's not going to be a common thing to do.

```suggestion
```

docs/getting_started.rst
```diff
@@ -75,3 +75,45 @@ Subscribe to tag changes
...
+* Query, create, modify, and delete Data Frame Tables and associated metadata
+
+* Query, query decimated, and append Data Frame Table data
```
spanglerco, line 98: Since we have two (or three) kinds of queries, maybe we can distinguish between them a little more. What do you think about something like this?

```suggestion
* Create and delete Data Frame Tables.

* Modify table metadata and query for tables by their metadata.

* Append rows of data to a table, query for rows of data from a table, and decimate table data.
```

It could also be useful to get Melissa's feedback for this one.

examples/dataframe/query_read_data.py
```diff
@@ -0,0 +1,12 @@
+from nisystemlink.clients.dataframe import DataFrameClient
+
+client = DataFrameClient()
+
+# List a table
+client.list_tables(take=1)
+
+# Get table metadata by table id
+client.get_table_metadata("639126ad256e5501f9a3cca5")
```
spanglerco, line 9: What if we grabbed the ID of the table that gets returned from `list_tables` above instead of hard-coding one for this example?

examples/dataframe/query_read_data.py
```diff
@@ -0,0 +1,12 @@
...
+
+# Get table data by table id
+client.get_table_data("639126ad256e5501f9a3cca5")
```
spanglerco, line 12: What if we showed building a decimated query instead? An example of getting undecimated data would probably want to show paging. Which I'm okay doing later once we have a helper for that.

examples/dataframe/delete_table.py
```diff
@@ -0,0 +1,6 @@
+from nisystemlink.clients.dataframe import DataFrameClient
+
+client = DataFrameClient()
+
+# Delete table by id
+client.delete_table("63912927256e5501f9a3cca6")
```
spanglerco, line 6: I think the API is going to be discoverable enough that maybe we can drop this example.

spanglerco (changes requested):

docs/getting_started.rst
cameronwaterman: It won't let me add Melissa to the review for some reason, but I'll leave this one open until I get her sign-off.

spanglerco (approved):

docs/getting_started.rst
```diff
@@ -75,3 +75,39 @@ Subscribe to tag changes
...
+Data Frame API
+-------
+
+Overview
+~~~~~~~~
+
+The :class:`.DataFrameClient` class is the primary entry point of the Data Frame API.
+
+When constructing a :class:`.DataFrameClient`, you can pass an
+:class:`.HttpConfiguration` (like one retrieved from the
```
MelissaHilliard, line 88: parenthetical statements can be difficult to localize. if that's a concern, consider rewriting

cameronwaterman: We don't currently localize these docs and don't have any immediate plans to so I think we can do without this change for now.

PR #45, opened 2023-03-16
feat: Add CSV export capability to DataFrameClient

kjohn1922:

Add `export_table_data` to the DataFrameClient, which calls the export route on the DataFrame Service API. 

Wrap the `response.iter_content` generator in a file-like object that can either be written to a file or loaded into a pandas dataframe, without loading the data into memory all at once beforehand.

Added an example of writing the export data to a file or pandas.

Parity of the python API with new features added to the DataFrame Service API.

Manual testing using Jupyter notebooks. Added unit tests for the generator wrapper class. Added an integration test for the export data method.


nisystemlink/clients/core/helpers/generator_file_like.py
kjohn1922, line 1: A couple of comments on this:
- It's returning the data as binary data. The binary can be written to a CSV file or to pandas directly, but if the user wants to  get the data as a string in code, they will need to decode it themselves. We did this in anticipation of adding binary export formats in the future (TDMS).
- I didn't implement `readline` or `readlines`, since they are not required for writing to a file or to pandas. I can add them if we want this to be a fully-functional file-like object, but they were getting a bit messy with iterating the binary and normalizing line endings to what python expects.

nisystemlink/clients/dataframe/_data_frame_client.py
```diff
@@ -21,7 +22,7 @@ def __init__(self, configuration: Optional[core.HttpConfiguration] = None):
                 is used.
 
         Raises:
-            ApiException: if unable to communicate with the Data Frame service.
+            ApiException: if unable to communicate with the DataFrame Service.
```
kjohn1922, line 25: Sneaking in these doc changes, after a conversation with Mark and Paul:
It's the DataFrame (one word) Service (capitalized), but the tables are data tables, not DataFrame tables.

cameronwaterman (approved):

kjohn1922:
I'm working through some type errors.

nisystemlink/clients/dataframe/_data_frame_client.py
```diff
@@ -275,14 +277,38 @@ def query_decimated_data(
...
             The decimated table data.
 
         Raises:
-            ApiException: if unable to communicate with the Data Frame service
+            ApiException: if unable to communicate with the DataFrame Service
+                or provided an invalid argument.
+        """
+        ...
+
+    @response_handler()  # type: ignore
```
kjohn1922, line 292: I get a mypy error here that I haven't figured out how to resolve:
````
PS C:\dev\ni\nisystemlink-clients-python> poetry run poe types
Poe => mypy --config-file mypy.ini nisystemlink examples tests
nisystemlink\clients\dataframe\_data_frame_client.py:293: error: Untyped decorator makes function "iter_content_filelike_wrapper" untyped
nisystemlink\clients\dataframe\_data_frame_client.py:297: error: Untyped decorator makes function "export_table_data" untyped
```

kjohn1922: We nee to suppress type checking for this, because we don't have the ability to add typing for response_handler.

nisystemlink/clients/core/helpers/iterator_file_like.py
```diff
@@ -0,0 +1,30 @@
+"""A file-like object adapter that wraps a python iterator, providing a way to
```
spanglerco, line 1: Should this documentation go under the class instead of the module?

nisystemlink/clients/dataframe/_data_frame_client.py
spanglerco: Should that be a comment in the code? :)

nisystemlink/clients/dataframe/_data_frame_client.py
```diff
@@ -275,14 +277,38 @@ def query_decimated_data(
...
+    def iter_content_filelike_wrapper(response: Response) -> IteratorFileLike:
```
spanglerco, line 293: Convention is for "private" methods to start with an underscore.

spanglerco (changes requested):

docs/api_reference/dataframe.rst
```diff
@@ -20,6 +20,7 @@ nisystemlink.clients.dataframe
    .. automethod:: append_table_data
    .. automethod:: query_table_data
    .. automethod:: query_decimated_data
+   .. automethod:: export_table_data
```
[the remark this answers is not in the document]
kjohn1922: If we move `export_table_data` above `query_decimated_data`, the rest of the order matches what we have in the Swagger UI. I agree that modify, delete, delete, modify feels inconsistent, though. Do we want to change the order for Swagger as well?

nisystemlink/clients/dataframe/_data_frame_client.py
```diff
@@ -275,14 +277,38 @@ def query_decimated_data(
...
+        return IteratorFileLike(response.iter_content(chunk_size=4096))
+
+    @iter_content_filelike_wrapper  # type: ignore
+    @post("tables/{id}/export-data", args=[Path, Body])
+    def export_table_data(
+        self, id: str, query: models.ExportTableDataRequest
+    ) -> IteratorFileLike:
```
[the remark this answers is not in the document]
kjohn1922: Added helpers to the documentation from the core.rst file, confirmed that IteratorFileLike appears in the documentation and that the return type gets rendered as a link.

nisystemlink/clients/core/helpers/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .iterator_file_like import IteratorFileLike
```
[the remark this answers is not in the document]
kjohn1922: Thanks for the explanation! I added an underscore to iterator_file_like and changed the imports.

nisystemlink/clients/dataframe/_data_frame_client.py
kjohn1922: Done :)

docs/api_reference/dataframe.rst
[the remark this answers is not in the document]
kjohn1922: I made the order match the swagger UI. If you're interested in reordering them in both places, let me know and I can make a quick tech debt item for that.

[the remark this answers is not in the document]
spanglerco: Probably matching the swagger UI is good enough for now.

spanglerco (approved):

mure:
From https://dev.azure.com/ni/DevCentral/_workitems/edit/2347097,

If the server terminates the connection for any reason, the client will get an unhelpful `requests.exceptions.ChunkedEncodingError` error. It would be nice to catch this and have a more user friendly message.

nisystemlink/clients/dataframe/_data_frame_client.py
mure: I added wrappers for the Uplink decorators here: https://github.com/ni/nisystemlink-clients-python/blob/master/nisystemlink/clients/core/_uplink/_methods.py
to add proper type annotations (and some default args). We could add `response_handler` there if we wanted to. It just pushes the `type: ignore` to a different module, though. My goal was to make the client implementation as smooth as possible. It's probably less important for this narrow use case.

pyproject.toml
```diff
@@ -44,6 +44,8 @@ flake8-import-order = "^0.18.1"
 pytest              = "^7.2.0"
 pytest-asyncio      = "^0.20.1"
 mypy                = "^0.982"
+pandas              = "^1.5.3"
```
mure, line 47: I'm not sure how I feel about including pandas (even as a "dev" dependency) when it's only used in an example. My preference would be to tell mypy to ignore it completely and then revert these changes.
We can add

```
[mypy-pandas.*]
ignore_missing_imports = True
```
to the `mypy.ini` file.

mure (changes requested):

nisystemlink/clients/dataframe/_data_frame_client.py
mure: Actually, we definitely need to bolster the types somehow. The function signature and docstring are getting lost in translation for the export method:

https://user-images.githubusercontent.com/9257800/228989125-0e868847-2c4f-4dad-acf8-bd37e8dd8237.mov

mure: I think we just need to invert the `response_handler` usage:

```python
def _iter_content_filelike_wrapper(response: Response) -> IteratorFileLike:
    return IteratorFileLike(response.iter_content(chunk_size=4096))

@response_handler(_iter_content_filelike_wrapper) # type: ignore
@post("tables/{id}/export-data", args=[Path, Body])
def export_table_data(
    self, id: str, query: models.ExportTableDataRequest
) -> IteratorFileLike: ...
```

Now my intellisense is as expected:
Screenshot 2023-03-30 at 6 57 09 PM

tests/core/test_generator_file_like.py
```diff
@@ -0,0 +1,34 @@
+from nisystemlink.clients.core.helpers import IteratorFileLike
```
mure, line 1: nit: this filename doesn't match the tested class anymore

tests/core/test_generator_file_like.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+class TestIteratorFileLike:
+    def test__negative_size__read__reads_all_data(self):
+        def generate_data():
+            yield b"123"
+            yield b"456"
+            yield b"789"
+
+        iterator_file_like = IteratorFileLike(generate_data())
```
mure, line 11: nit: We could simplify this to something like:

```python
iterator_file_like = IteratorFileLike(iter([b"123", b"456", b"789"]))
```

nisystemlink/clients/dataframe/_data_frame_client.py
kjohn1922: Thanks for the information! I added a response_handler wrapper to _methods and inverted the response handler usage.

pyproject.toml
kjohn1922: I made this change, but I'm still getting a type error in the example file, for the argument passed to read_csv.
Right now, I have the pandas lines commented out... which is not unreasonable, because if you run the example as-is you get an error because the iterator is already exhausted by the "save to file" part of the example.
If we don't like it commented out, though, I can look into the error more later, since I need to help with testing today.

mure: I'm fine with having that part of the example commented out. I think it's pretty common to see stuff like that in code snippets.

spanglerco (approved):

mypy.ini
```diff
@@ -11,6 +11,9 @@ disallow_untyped_decorators=True
 
 strict_equality=True
 
+[mypy-pandas.*]
+ignore_missing_imports=True
+
```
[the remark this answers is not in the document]
kjohn1922: Ah, good point. Removed.

PR #51, opened 2023-11-03
feat: minimal implementation of Auth and WebApp clients

santhoshramaraj:

Adds support for Auth and WebApp services with limited scope

Auth and WebApp clients are a dependency for DataSpace client

Tested with main-test using Pytest


mure:
FYI I updated your PR title to match the [commit format needed](https://github.com/angular/angular.js/blob/master/DEVELOPERS.md#commits) for our automated release workflow.

nisystemlink/clients/auth/models/_additional_models.py
```diff
@@ -0,0 +1,25 @@
...
+
+from typing import Dict, List, Optional
+
+from nisystemlink.clients.core._uplink._json_model import JsonModel
+from pydantic import Field
+
+from ._auth_models import AuthPolicy, Error, Org, User, Workspace
+
+
+class AuthResponse(JsonModel):
```
mure, line 11: My preference is to avoid using the 'Response' suffix on models. There's a few arguments against it, such as:
- It would be redundant to put it on everything
- Doesn't convey any useful meaning to users

Couple ideas are `AuthResult` or `AuthInfo`, but I'm open to alternatives.

nisystemlink/clients/auth/models/__init__.py
```diff
@@ -0,0 +1,3 @@
+from ._additional_models import AuthResponse
```
mure, line 1: Is the intention here to only re-export the models that the minimal implementation of the client actually uses? That makes sense to me.

nisystemlink/clients/auth/models/_additional_models.py
```diff
@@ -0,0 +1,25 @@
...
+    """Details of authenticated caller"""
+    org: Optional[Org]
+    """Organization of authenticated caller"""
+    workspaces: Optional[List[Workspace]]
+    """List of workspaces the authenticated caller has access"""
+    policies: Optional[List[AuthPolicy]]
+    """List of policies for the authenticated caller"""
+    properties: Optional[Dict[str, str]] = Field(None, example={"key1": "value1"})
+    """A map of key value properties"""
+    error: Optional[Error]
```
mure, line 24: It doesn't make sense to have the error field here. The HTTP API will return either the auth information OR an error, but not both in the same JSON body.

nisystemlink/clients/auth/models/_auth_models.py
mure, line 1: Does the model generator have a configuration option that lets us split each model into its own file? I think we should do that if possible.

nisystemlink/clients/auth/models/_auth_models.py
```diff
@@ -0,0 +1,373 @@
...
+    """
+    Whether the workspace is enabled or not
+    """
+    default: Optional[bool] = Field(None, example=True)
+    """
+    Whether the workspace is the default. The default workspace is used when callers omit a workspace id
+    """
+
+
+class Error(JsonModel):
```
mure, line 268: We've already created the error model in the core package: https://github.com/ni/nisystemlink-clients-python/blob/master/nisystemlink/clients/core/_api_error.py
so this is unnecessary.

nisystemlink/clients/auth/models/_auth_models.py
```diff
@@ -0,0 +1,373 @@
...
+    niua_id: Optional[str] = Field(None, alias="niuaId", example="example@email.com")
+    """
+    The external id (niuaId, SID, login name)
+    """
+    login: Optional[str] = None
+    """
+    The login name of the user. This the "username" or equivalent entered when
+    the user authenticates with the identity provider.
+    """
+    accepted_to_s: Optional[bool] = Field(None, alias="acceptedToS", example=True)
```
mure, line 196: I know we're not using this model for anything yet, but `accepted_to_s` is a weird name. Probably should just be `accepted_tos`

nisystemlink/clients/auth/models/_auth_models.py
```diff
@@ -0,0 +1,373 @@
...
+class AuthStatement(JsonModel):
+    actions: Optional[List[str]] = None
+    """
+    A list of actions the user is allowed to perform
+    """
+    resource: Optional[List[str]] = None
+    """
+    A list of resources the user is allowed to access
+    """
+    workspace: Optional[str] = Field(None, example="workspace-id")
```
mure, line 23: Do you know if the `example` parameter actually does anything in Pydantic?

nisystemlink/clients/webapp/_webapp_client.py
```diff
@@ -0,0 +1,51 @@
...
+    def query_webapps(
+        self, query: models.WebAppsAdvancedQuery
+    ) -> models.WebAppsResponse:
+        """Use the Dynamic Linq query language to specify filters for webapps.
+        An empty request body queries all webapps.
+        """
+        ...
+
+    @get("{id}/content")
+    def get_content(self, id: str) -> Any:
```
mure, line 41: It's been awhile since I looked at the uplink docs, but with the `Any` type this will return the whole `requests.Response` object, right? We don't do that anywhere else. Maybe we can reuse the `IteratorFileLike` object that we created for the export method of the DFS client?https://github.com/ni/nisystemlink-clients-python/blob/3459a508e43512b4e747d509c3ba58fd4c8f9aa5/nisystemlink/clients/dataframe/_data_frame_client.py#L305

nisystemlink/clients/webapp/models/_additional_models.py
```diff
@@ -0,0 +1,26 @@
...
+    List of webapps.
+    """
+    continuation_token: Optional[str] = Field(
+        None, alias="continuationToken", example="token"
+    )
+    """
+    The continuation token can be used to paginate through the webapp query results.
+    Provide this token in the next query webapps call.
+    """
+    error: Optional[Error]
```
mure, line 23: Same as above comment - we shouldn't have the error here.

tests/integration/auth/test_auth.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+@pytest.mark.enterprise
+@pytest.mark.integration
+class TestAuth:
+    def test__get_auth__returns(self, client: AuthClient):
+        auth_response = client.get_auth()
+        assert getattr(auth_response, "error") is None
+        assert auth_response.workspaces is not None
+        assert len(auth_response.workspaces) >= 1
```
mure, line 18: It would be better if this validated more of the fields in the response. Although it will admittedly be a bit of work to set up. We'd need the test to:
- Create a new user
- Generate an API key for that user
- Create an auth client with the new configuration
- Then call `.get_auth()`and assert the fields of the created user

I think this is something we should do, but I'm okay with having it in a separate PR that follows this one.

mure (changes requested):

nisystemlink/clients/auth/_auth_client.py
```diff
@@ -0,0 +1,33 @@
...
+        Raises:
+            ApiException: if unable to communicate with the Auth Service.
+        """
+        if configuration is None:
+            configuration = core.JupyterHttpConfiguration()
+
+        super().__init__(configuration, base_path="/niauth/v1/")
+
+    @get("auth")
+    def get_auth(self) -> models.AuthResponse:
```
mure, line 31: Something else I thought of: the only auth functionality we need for data spaces is to get workspaces by name, right? While this method could be useful, you could just call `/niuser/v1/workspaces` instead, which would have a lower testing burden.

nisystemlink/clients/auth/_auth_client.py
```diff
@@ -0,0 +1,33 @@
...
+from typing import Optional
+
+from nisystemlink.clients import core
+from nisystemlink.clients.core._uplink._base_client import BaseClient
+from nisystemlink.clients.core._uplink._methods import get
+
+from . import models
+
+
+class AuthClient(BaseClient):
```
mure, line 12: I think it's very likely we'll want to combine the /niuser and /niauth methods into a single client. There's not really a meaningful distinction between them and they're both implemented in the same service anyway. With that in mind, is `AuthClient` the name we want to go with? I could see `UserClient`, `AdminClient`, `AccessControlClient` off the top of my head.

@spanglerco, @kjohn1922 , @cameronwaterman any opinions?

nisystemlink/clients/auth/models/_additional_models.py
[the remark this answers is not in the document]
santhoshramaraj: Yes, these were models that were not defined in the OpenAPI yaml and hence manually created in a separate file to keep it separate.

nisystemlink/clients/auth/models/_additional_models.py
santhoshramaraj: The term 'Response' was used to be consistent with the OpenAPI yaml. I am good with changing it to 'AuthInfo' for this instance.
image

nisystemlink/clients/auth/models/__init__.py
santhoshramaraj: Yes, and consistent with Dataframe client models. On the other hand, exporting everything has a side effect of adding the JsonModel and Field classes to be included in the readthedocs automated generation.

nisystemlink/clients/auth/models/_auth_models.py
santhoshramaraj: These pydantic models follow the OpenApi yaml 1-1 and hence each model ends up with its copy of the API Error model as defined in the yaml.

Though logically correct to use the same Error model, it would not be practical to see what other models are common across clients and move them to the core for reuse.

nisystemlink/clients/auth/_auth_client.py
santhoshramaraj: It may not fit the purposes of DataSpaces as per another internal discussion.

nisystemlink/clients/auth/models/_auth_models.py
santhoshramaraj: Useful only when exporting the Pydantic model as JSON Schema and hence not useful in the scope of the nisystemlink-clients module.

nisystemlink/clients/auth/_auth_client.py
mure: I talked to @spanglerco and we agree that the `/auth` route is the correct way to get available workspaces.

Unrelated, I think I would prefer this method to be called `authenticate()` rather than `get_auth()`.

nisystemlink/clients/auth/models/_auth_models.py
santhoshramaraj: The data model generator does not have the option to generate models into separate files.

Generating each model in its own file will be messy unless logically grouped. Storing the model in a single file keeps it simple to separate the generated model vs the manually created ones.

nisystemlink/clients/auth/models/_auth_models.py
mure: My thinking is that we should use the generated models as a starting point to save time, but not try to keep them perfectly in sync with the Swagger doc. We should be willing to make manual changes, such as consolidating duplicates and modifying the docstrings if needed to match specifics of the Python implementation. 

For example, here's an instance where we had to change `null` to `None`:

https://github.com/ni/nisystemlink-clients-python/blob/3459a508e43512b4e747d509c3ba58fd4c8f9aa5/nisystemlink/clients/dataframe/models/_modify_table_request.py#L33

mure (changes requested):

nisystemlink/clients/auth/models/_auth_info.py
```diff
@@ -0,0 +1,23 @@
...
+    user: Optional[User]
+    """Details of authenticated caller"""
+    org: Optional[Org]
+    """Organization of authenticated caller"""
+    workspaces: Optional[List[Workspace]]
+    """List of workspaces the authenticated caller has access"""
+    policies: Optional[List[AuthPolicy]]
+    """List of policies for the authenticated caller"""
+    properties: Optional[Dict[str, str]] = Field(None, example={"key1": "value1"})
+    """A map of key value properties"""
```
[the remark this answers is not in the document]
santhoshramaraj: It looks like the Swagger doc has modeled these as optional fields.

https://test-api.lifecyclesolutions.ni.com/niapis/?urls.primaryName=Auth%20Service
https://test-api.lifecyclesolutions.ni.com/niauth/swagger/v1/niauth.yaml

```
  AuthResponse:
    description: Auth Response
    schema:
      type: object
      title: Auth Response
      properties:
        user:
          $ref: '#/definitions/User'
        org:
          $ref: '#/definitions/Org'
        workspaces:
          type: array
          items:
            $ref: '#/definitions/Workspace'
        policies:
          type: array
          items:
            $ref: '#/definitions/AuthPolicy'
        properties:
          type: object
          description: A map of key value properties
          additionalProperties:
            type: string
          example:
            key1: value1
```

santhoshramaraj:
Close stale PR

PR #52, opened 2024-02-06
Initial commit of genericized API

ChrisStrykesAgain:

This defines the basics of a possible cross-service API that allows thinking in more test- and data-centric terms, and prevents having to think about which individual service one needs to be interfacing with.

At least one customer has found it useful as-is, however the purpose of this PR is to refine it (hopefully greatly). I think there is still substantial work to be done before we can get it to that point, but my hope is that discussion on the PR will help push it in that direction.

TODO: Define tests

### DIFF
--- a/nisystemlink/cross-service/GenericApi.py
+++ b/nisystemlink/cross-service/GenericApi.py
@@ -1,0 +1,1040 @@
+"""This is a module for a cross-service API class."""
+import asyncio
+import datetime
+import http.client
+import json
+import os
+import time
+import urllib.parse
+from collections import defaultdict
+from enum import Enum
+from io import BytesIO
+from typing import Any, Dict, List, Tuple  # noqa: F401
+
+
+import aiohttp
+import pandas as pd
+import systemlink.clients.nifile.api.files_api as files_api
+import systemlink.clients.nifile.models as fis_models
+import systemlink.clients.nitestmonitor.api.results_api as results_api
+import systemlink.clients.nitestmonitor.models as tm_models
+from nxtdms import TdmsFile
+
+
+class DatatableType(str, Enum):
+    """Enum to describe the type of data in a table."""
+
+    RAW = "raw"
+    CYCLE = "cycle"
+    CALCULATED = "calculated"
+    DATASHEET = "datasheet"
+
+
+class CrossServiceApi:
+    """This is a class to abstract the SystemLink and related APIs.
+
+    This will make it more conducive to usage by broader users.
+
+    Attributes
+    ----------
+    dataHandle : str
+        The handle for the Data API instance
+    tablesHandle : str
+        The handle ofr the Tables API instance
+    id : string
+        the alphanumeric string representing the SystemLink id of
+        the test result in question
+
+    Methods
+    -------
+    findOrCreateRecord(self, identifier: str) -> str:
+        Either finds a record with a name that matches the identifier or creates
+        a new reord with that name, the sets the active record for the object
+        to the ID of that record.
+    setFileMeta(self, fileMeta: Dict[str, str]):
+        Sets the File Metadata of the active result.
+    setChannelMeta(self, channel: str, channelMeta: Dict[str, str]):
+        Sets the metadata for the specified channel.
+    setChannelMapping(self, channelMappings: Dict[str, str]):
+        Sets the channel mappings of a given file. This should map the
+        file-centric names to the more commonly used aliases which
+        are referenced elsewhere.
+    appendData(self, data: pd.DataFrame):
+        Appends data to a table.
+    readData(self, channels: List[str]) -> pd.DataFrame:
+        Reads data from a table.
+    getTestMeta(self, key: List[str] = None) -> Dict[str, str]:
+        Gets some or all fo the metadata associated with a test.
+    getChannelMeta(self, channel: List[str] = None, key: List[str] = None) -> Dict[str, Dict[str, str]]: # noqa: E501
+        Returns some or all metadata associated with some or all of the channels in a given test.
+    """
+
+    def __init__(
+        self,
+        result_handle: results_api.ResultsApi = None,
+        fis_handle: files_api.FilesApi = None,
+        sl_api_key: str = None,
+        sl_uri: str = None,
+    ):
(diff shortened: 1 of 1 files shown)


nisystemlink/cross-service/GenericApi.py
```diff
@@ -0,0 +1,1040 @@
...
+                raise InternalError("Record has no dataframe attached")
+            for target_frame in test_meta["data_table_ids"]:
+                specific_frame_name = self._get_datatable_name(target_frame)
+                if source_frame == specific_frame_name:
+                    target_to_source_mapping[target_frame] = source_frame
+            if source_frame not in target_to_source_mapping.values():
+                raise KeyError("No corresponding target frame.")
+        return target_to_source_mapping
+
+    async def append_data(
```
mure, line 325: `append_data` writes data to multiple tables on a test result from a Pandas df. Interacting with the DFS using Pandas is probably the most useful thing we could add to this package. I could see us having two separate classes for reading and writing, something like `RowDataReader` and `RowDataWriter`. Usage would be something like:

```py
with RowDataWriter(chunk_size=5000, end_of_data_on_exit=True) as writer:
  # creates tables matching Pandas columns and appends data in batches
  writer.write_df('table1', df1)
  writer.write_df('table2', df2)
```

nisystemlink/cross-service/GenericApi.py
```diff
@@ -0,0 +1,1040 @@
...
+            dataresponse = json.loads(payload[2:-1])
+            actual_row_count = dataresponse["totalRowCount"]
+            if actual_row_count == row_count:
+                all_there = True
+            else:
+                num_retries += 1
+                time.sleep(30)
+        return all_there
+
+    async def read_data(
```
mure, line 442: As mentioned in the previous comment, we could implement a new `RowDataReader` class. The [export method](https://github.com/ni/nisystemlink-clients-python/blob/master/examples/dataframe/export_data.py) that already exists also lets you stream a table into a Pandas DF.

nisystemlink/cross-service/GenericApi.py
```diff
@@ -0,0 +1,1040 @@
...
+            headers=self._headers,
+        )
+        response = http_conn.getresponse()
+        payload = str(response.read()).replace("\\", "")
+        dataresponse = json.loads(payload[2:-1])
+        for column in dataresponse["columns"]:
+            column_names.append(column["name"])
+        return column_names
+
+    async def get_test_meta(self, record_id: str, keys: List[str] = None) -> Dict[str, str]:
```
mure, line 656: Not needed if we have a test monitor client

nisystemlink/cross-service/GenericApi.py
```diff
@@ -0,0 +1,1040 @@
...
+                headers=self._headers,
+            )
+            response = http_conn.getresponse()
+            payload = str(response.read()).replace("\\", "")
+            dataresponse = json.loads(payload[2:-1])
+            for column in dataresponse["columns"]:
+                column_names.append(column["name"])
+        return column_names
+
+    async def get_specific_columns(self, table_id: str) -> List[str]:
```
mure, line 637: This can easily be done with the current DataFrameClient

nisystemlink/cross-service/GenericApi.py
```diff
@@ -0,0 +1,1040 @@
...
+                        return_dict[property] = resp["properties"][property]
+                elif key == "status":
+                    return_dict["status"] = resp["status"]["status_name"]
+                elif key in ["status_type_summary", "id", "file_ids", "workspace"]:
+                    continue
+                else:
+                    return_dict[key] = resp[key]
+        return return_dict
+
+    async def get_channel_name_by_synonym(
```
mure, line 718: This method seems to be based on a custom `synonym` property. Since it's not a first-class part of the SystemLink API, behavior like this should live in a different package (in my opinion)

nisystemlink/cross-service/GenericApi.py
```diff
@@ -0,0 +1,1040 @@
...
+
+    async def get_test_id_from_file_id(self, file_id: str):
+        """Get the test id that corresponds to a given file."""
+        response = fis_models.QueryResponse(
+            await self._fis_handle.list_available_files_get(id=file_id)
+        )
+        files = response.links.available_files
+        return files[0].properties["test_id"]
+
+    async def create_datatable(
```
mure, line 836: This is already implemented by the Data Frame Client

nisystemlink/cross-service/GenericApi.py
```diff
@@ -0,0 +1,1040 @@
...
+            if r.status == 204:
+                upload_success = True
+            else:
+                num_trys += 1
+                print(r.status)
+                print("Upload attempt {} failed".format(num_trys))
+        local_http_conn.close()
+        return upload_success
+
+    async def get_test_id_from_file_id(self, file_id: str):
```
mure, line 828: This gets a hard-coded property from a file - don't think it should be in this package

nisystemlink/cross-service/GenericApi.py
```diff
@@ -0,0 +1,1040 @@
...
+            "PATCH", f"/nidataframe/v1/tables/{datatable}", body=body, headers=self._headers
+        )
+        response = self._http_conn.getresponse()
+        if response.status >= 400:
+            raise InternalError(response.read())
+        else:
+            response.read()
+        return datatable
+
+    async def get_workspace_id(self, desired_workspace: str) -> str:
```
mure, line 868: At some point we'll have an AuthClient (name TBD) that implements this

nisystemlink/cross-service/GenericApi.py
```diff
@@ -0,0 +1,1040 @@
...
+        if response.status >= 400:
+            raise InternalError(response.read())
+        payload = ""
+        while chunk := response.read(200):
+            payload += repr(chunk)
+        query_response = json.loads(payload.replace("'", '"')[2:-1])
+        workspace_id = query_response["workspaces"][0]["id"]
+        return workspace_id
+
+    async def get_specific_cycle_table_id(
```
mure, line 884: This and the methods below are all based on the existence of "proprietary" custom properties, so they should be in a separate package

PR #61, opened 2024-06-12
feat: Make DataFrame and Spec clients compatible with SystemLink Client http configuration

santhoshramaraj:

Add `JupyterHttpConfiguration` as a fallback configuration for the `HttpConfigurationManager`. Allows the Clients to authenticate agnostic of the environment they're run from.

Enables SystemLink clients to adapt to supported run environments when using the default value for `configuration` arg like `client = DataFrameClient()`

Manual testing

- Run https://github.com/ni/nisystemlink-clients-python/blob/master/examples/dataframe/export_data.py on the computer with SystemLink Client connected with an SLE server
- `KeyError` occurs because `JupyterHttpConfiguration` is not available outside SLE
- Update the `DataFrameClient()` to use `core.HttpConfigurationManager.get_configuration()` as the default configuration
- Run the example, now, there should be no `KeyError` signifying that the client was created using the SystemLink client's HTTP Credentials

### DIFF
--- a/nisystemlink/clients/core/_http_configuration_manager.py
+++ b/nisystemlink/clients/core/_http_configuration_manager.py
@@ -23,10 +23,8 @@ class HttpConfigurationManager:
     HTTP_LOCALHOST_CONFIGURATION_ID = "SYSTEMLINK_LOCALHOST"
     """The default ID of the SystemLink Server's configuration on the SystemLink Server itself."""
 
-    JUPYTER_HUB_CONFIGURATION_ID = "SYSTEMLINK_JUPYTER_HUB"
-    """ID for the http configuration for Jupyter Notebooks run in a SystemLink Enterprise environment."""
-
     _configs = None
+    _juypter_config = None
 
     @classmethod
     def get_configuration(
@@ -81,15 +79,22 @@ class HttpConfigurationManager:
         """
         if cls._configs is None:
             cls._configs = cls._read_configurations()
+
+        if cls._juypter_config is None:
+            try:
+                cls._juypter_config = core.JupyterHttpConfiguration()
+            except KeyError:
+                # Env variables for Jupyter Hub execution are not available.
+                pass
+
         master_config = cls._configs.get(cls.HTTP_MASTER_CONFIGURATION_ID)
         if master_config is not None:
             return master_config
         localhost_config = cls._configs.get(cls.HTTP_LOCALHOST_CONFIGURATION_ID)
         if localhost_config is not None:
             return localhost_config
-        jupyter_config = cls._configs.get(cls.JUPYTER_HUB_CONFIGURATION_ID)
-        if jupyter_config is not None:
-            return jupyter_config
+        if cls._juypter_config is not None:
+            return cls._juypter_config
         return None
 
     @classmethod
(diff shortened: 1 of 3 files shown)


nisystemlink/clients/core/_http_configuration_manager.py
```diff
@@ -23,6 +23,9 @@ class HttpConfigurationManager:
     HTTP_LOCALHOST_CONFIGURATION_ID = "SYSTEMLINK_LOCALHOST"
     """The default ID of the SystemLink Server's configuration on the SystemLink Server itself."""
 
+    JUPYTER_HUB_CONFIGURATION_ID = "SYSTEMLINK_JUPYTER_HUB"
```
spanglerco, line 26: Since this is a "virtual" configuration that doesn't actually exist on disk, I wonder if we should track it separately than in the `_configs` dict. If someone made a custom configuration with this ID (i.e., created a file for it in `C:\ProgramData\National Instruments\Skyline\HttpConfigurations`), it'll overwrite the Juypter one _and_ get treated as a fallback, which custom configurations don't normally do.

Though it's probably reasonable to treat any configurations starting with `SYSTEMLINK_` as "reserved" and not something customers should create themselves, it could be nice to avoid this ambiguity.

For example, we could have a separate `_juypter_config` attribute next to `_configs`.

spanglerco (changes requested):

nisystemlink/clients/core/_http_configuration_manager.py
santhoshramaraj: It makes sense that `id` based configurations are limited to physical configurations read from disk and Jupyter config is special.

nisystemlink/clients/core/_http_configuration_manager.py
```diff
@@ -78,12 +79,22 @@ def _fallback(cls) -> Optional[core.HttpConfiguration]:
         """
         if cls._configs is None:
             cls._configs = cls._read_configurations()
+
+        if cls._juypter_config is None:
```
spanglerco, line 83: Can we combine these blocks so we only try to load the Jupyter config once even if it fails?

```suggestion
```

spanglerco (approved):

nisystemlink/clients/core/_http_configuration_manager.py
santhoshramaraj: @spanglerco, is the concern that when running outside JupyterHub, the JupyterHttpConfig will be attempted every time a client creates and obtains the HttpConfiguration?

spanglerco: Yes

nisystemlink/clients/core/_http_configuration_manager.py
```diff
@@ -78,12 +79,22 @@ def _fallback(cls) -> Optional[core.HttpConfiguration]:
...
+            try:
+                cls._juypter_config = core.JupyterHttpConfiguration()
+            except KeyError:
+                # Env variables for Jupyter Hub execution are not available.
```
mure, line 87: Extreme nitpick, but technically this configuration is for notebook execution in general, which isn't always through JupyterHub.

```suggestion
                # Env variables for Jupyter notebook execution are not available.
```

mure (approved):
Thanks for doing this!

nisystemlink/clients/core/_http_configuration_manager.py
santhoshramaraj: Refactored into a `cls._virtual_config` to load once like the file-based configs.

spanglerco (approved):

PR #65, opened 2024-07-01
feat: Add Client for File Service

santhoshramaraj:

- Add Client for the File Service
- Add support for basic File Service APIs
- Add a basic back-off-retry mechanism for FileClient 
   ```py
   @retry(when=retry.when.status(429), stop=retry.stop.after_attempt(5))
   class FileClient(BaseClient):
   ```
   Uplink uses exponential backoff by default.

Improve Developer Experience by using `nisystemlink-clients` instead of `systemlink-sdk`, which is installed with the SystemLink Client application.

- Integration tests were implemented and tested against the test tier.
- Manual testing of back-off-retry logic by commenting out the logic and run integration tests on dev machine against test tier.


examples/file/download_file.py
```diff
@@ -0,0 +1,30 @@
...
+if not files.available_files:
+    raise Exception(f"File ID {file_id} not found.")
+
+
+file_name = "Untitled"
+
+file_properties = files.available_files[0].properties
+
+if file_properties:
+    file_name = file_properties.get("Name", "Untitled")
```
adamarnesen, line 23: I don't personally like the `.get` method on dictionaries. Can we use `file_properties["Name"]` instead?

nisystemlink/clients/file/_file_client.py
```diff
@@ -0,0 +1,177 @@
...
+            and descending if True. Defaults to False.
+            file_ids: Comma-separated list of file IDs to search by. Defaults to None.
+
+        Returns:
+            File Query Response
+
+        Raises:
+            ApiException: if unable to communicate with the File Service.
+        """
+        # Uplink does not support enum serializing into str
```
adamarnesen, line 119: This is annoying huh? I'm ok with this workaround for now. I suspect there may be other workarounds, but I'm for getting this API in soon even if it needs to have this workaround.

nisystemlink/clients/file/models/_file_metadata.py
```diff
@@ -0,0 +1,65 @@
...
+class FileMetadata(JsonModel):
+    field_links: Optional[Dict[str, Link]] = Field(None, alias="_links")
+    """
+    The links to access and manipulate the file:
+    - data: Link to download the file using a GET request
+    - delete: Link to delete the file using a DELETE request
+    - self: Link to the file's metadata
+    - updateMetadata: Link to update the file's metadata using a POST request
+    """
+    created: Optional[datetime] = Field(None, example="2018-05-15T18:54:27.519Z")
```
adamarnesen, line 19: I think our style guide recommends having spaces between fields. I would expect to have a blank line before each here.

nisystemlink/clients/file/models/_file_metadata.py
```diff
@@ -0,0 +1,65 @@
...
+    properties: Optional[Dict[str, str]] = Field(None)
+    """
+    The file's properties
+    Example - {"Name": "myfile.txt", "MyProperty": "Value"}
+    """
+    meta_data_revision: Optional[int] = Field(None, alias="metaDataRevision", example=1)
+    """
+    The file's properties revision number. When a file is uploaded, the revision number starts at 1.
+    Every time metadata is updated, the revision number is incremented by 1.
+    New in version 3 of the listFiles operation.
```
adamarnesen, line 41: Probably don't need this version-specific doc string.

tests/integration/file/test_file_client.py
```diff
@@ -0,0 +1,92 @@
...
+FILE_NOT_FOUND_ERR = "Not Found"
+
+
+@pytest.fixture(scope="class")
+def client(enterprise_config) -> FileClient:
+    """Fixture to create a FileClient instance."""
+    return FileClient(enterprise_config)
+
+
+# @pytest.fixture(scope="class")
```
adamarnesen, line 21: We need to either get these fixtures working, or just remove them. Let's not include a bunch of commented out code.

tests/integration/file/test_file_client.py
```diff
@@ -0,0 +1,92 @@
...
+
+@pytest.mark.enterprise
+@pytest.mark.integration
+class TestFileClient:
+    def test__api_info__returns(self, client: FileClient):
+        api_info = client.api_info()
+        assert len(api_info.dict()) != 0
+
+    def test__get_files__succeeds(self, client: FileClient):
+        take_count = 1
```
adamarnesen, line 71: You probably need to have a setup here that at least uploads one file. There is no guarantee that there are _any_ files on the test instance. So, the test needs to make sure it sets up the conditions properly before testing them.

tests/integration/file/test_file_client.py
```diff
@@ -0,0 +1,92 @@
...
+        files = client.get_files(take=take_count)
+
+        assert len(files.available_files) == take_count
+        assert files.total_count >= take_count
+
+    def test__delete_file__invalid_id_raises(self, client: FileClient):
+        with pytest.raises(ApiException, match=FILE_NOT_FOUND_ERR):
+            client.delete_file(file_id=INVALID_FILE_ID, force=True)
```
adamarnesen, line 79: You may need to do  a pre-check that the ID is actually invalid. If I upload a files to the test system that matches `INVALID_FILE_ID` then I can make this test fail when it shouldn't

tests/integration/file/test_file_client.py
```diff
@@ -0,0 +1,92 @@
...
+
+    # def test__delete_file__succeeds(self, client: FileClient):
+    #     client.delete_file(file_id="7729547d-95af-4beb-a7a8-a913c0a23de4", force=True)
+
+    def test__download_file__invalid_id_raises(self, client: FileClient):
+        with pytest.raises(ApiException, match=FILE_NOT_FOUND_ERR):
+            client.download_file(file_id=INVALID_FILE_ID)
+
+    # def test__download_file__succeeds(self, client: FileClient, test_download_file: FileMetadata):
```
adamarnesen, line 88: Before we submit this, we need to at least have a working test for downloading a file.

adamarnesen (changes requested):

santhoshramaraj:
@adamarnesen refactored and added all key APIs for FileClient

adamarnesen (approved):

nisystemlink/clients/file/_file_client.py
```diff
@@ -0,0 +1,256 @@
...
+)
+from nisystemlink.clients.core.helpers import IteratorFileLike
+from requests.models import Response
+from uplink import Body, Part, Path, Query, retry
+
+
+from . import models
+
+
+def _iter_content_filelike_wrapper(response: Response) -> IteratorFileLike:
```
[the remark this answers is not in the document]
santhoshramaraj: Refactored under `nisystemlink.clients.core._uplink._file_like_response`

nisystemlink/clients/file/_file_client.py
mure: I don't know how to fix the bool serialization, though. Normally our services would be case insensitive for stuff like this.

nisystemlink/clients/file/_file_client.py
```diff
@@ -0,0 +1,253 @@
...
+
+        Args:
+            files: The description of files to delete.
+            force: Whether the deletion of files will be forced. Defaults to False.
+
+        Raises:
+            ApiException: if unable to communicate with the File Service.
+        """
+
+    # TBD: Error with poe types - Untyped decorator makes function "download_file" untyped
```
mure, line 165: nitpick, but we've used TODO instead of TBD in other places in this project

nisystemlink/clients/file/_file_client.py
```diff
@@ -0,0 +1,253 @@
...
+
+        Raises:
+            ApiException: if unable to communicate with the File Service.
+        """
+        if metadata:
+            metadata_str = json.dumps(metadata)
+        else:
+            metadata_str = None
+
+        resp = self.__upload_file(
```
mure, line 234: Could we parse the URI in the response and just return the string ID from this method?

nisystemlink/clients/file/models/_file_metadata.py
```diff
@@ -0,0 +1,74 @@
...
+    the user has associated privileges to view file paths.
+    """
+
+    properties: Optional[Dict[str, str]] = Field(None)
+    """
+    The file's properties
+    Example - {"Name": "myfile.txt", "MyProperty": "Value"}
+    """
+
+    meta_data_revision: Optional[int] = Field(None, alias="metaDataRevision", example=1)
```
mure, line 42: `metadata` isn't typically broken into two words

nisystemlink/clients/file/models/_operations.py
```diff
@@ -0,0 +1,38 @@
+from typing import Optional
+
+from nisystemlink.clients.core._uplink._json_model import JsonModel
+from pydantic import Field
+
+
+class Operation5(JsonModel):
```
mure, line 7: This model is already in the core package

nisystemlink/clients/file/utilities.py
```diff
@@ -0,0 +1,35 @@
...
+
+    Returns:
+        ID of the File extracted from the `uri`.
+    """
+    # Split the uri by '/' and get the last part
+    parts = uri.split("/")
+    return parts[-1]
+
+
+def rename_file(client: FileClient, file_id: str, name: str) -> None:
```
mure, line 23: Our clients have been mostly 1 to 1 with the HTTP API, but I think I'd rather this be a method on the FileClient itself instead of a separate function.

mure (approved):

cameronwaterman (approved):

nisystemlink/clients/file/_file_client.py
santhoshramaraj: Yes, I saw your TODO note and went with the wrapper method instead of Literal to keep it simple for the user anyway, I need to workaround the case sensitivity for boolean for that API.

nisystemlink/clients/file/models/_file_metadata.py
santhoshramaraj: It looks like the API implemented it as `metaDataRevision` and hence only `meta_data_revision` will result in a working automatic case conversion.

santhoshramaraj: It looks like the API implemented it as `metaDataRevision` and hence only `meta_data_revisio`n will result in a working automatic case conversion.

nisystemlink/clients/file/_file_client.py
```diff
@@ -0,0 +1,253 @@
...
+
+        return resp
+
+    @delete("service-groups/Default/files/{id}", args=[Path, Query])
+    def delete_file(self, id: str, force: bool = False) -> None:
+        """Deletes the file indicated by the `file_id`.
+
+        Args:
+            id: The ID of the file.
+            force: Whether the deletion of a file will be forced. Defaults to False.
```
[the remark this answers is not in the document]
santhoshramaraj: Force delete is required when the file may no longer exist at the expected location of the file service, which can cause an error message if you attempt to delete it.

It looks like always setting `force = True` shouldn't hurt.

nisystemlink/clients/file/models/_file_query_response.py
```diff
@@ -0,0 +1,28 @@
...
+from pydantic import Field
+
+from ._file_metadata import FileMetadata
+from ._link import Link
+
+
+class FileQueryResponse(JsonModel):
+    """The result of a file query"""
+
+    field_links: Dict[str, Link] = Field(alias="_links")
```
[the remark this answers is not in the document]
santhoshramaraj: The bottleneck to leave out `_links` is that the `conftest.py` forbids extra fields in the response compared to the Pydantic model.

https://github.com/ni/nisystemlink-clients-python/blob/8bc8f6309fd44294fc1f68fc6cca818ee7ec736d/conftest.py#L91

PR #69, opened 2024-08-08
feat: Add client for SystemLink products API

adamarnesen:

This PR adds client methods for all of the `nitestmonitor` product endpoints as well as testing for these endpoints.

This is an enhancement to the existing test monitor client and will enable users to more easily interact with SystemLink products. 


nisystemlink/clients/testmonitor/_test_monitor_client.py
```diff
@@ -23,6 +26,159 @@ def api_info(self) -> models.ApiInfo:
...
+        Returns:
+            A paged list of products with a continuation token to get the next page.
+
+        Raises:
+            ApiException: if unable to communicate with the ``/nitestmonitor`` Service or provided invalid
+                arguments.
+        """
+        ...
+
+    @returns.json
```
adamarnesen, line 112: @mure @cameronwaterman 
I'm struggling on getting `poe types` to be happy with this. I keep getting `error: Untyped decorator makes function "query_product_values" untyped`
Any suggestions?

adamarnesen: I ended up just marking this as ignore. If anyone has other suggestions, I'm open to it.

nisystemlink/clients/testmonitor/_test_monitor_client.py
santhoshramaraj, line 1: Please update the `__init__()` to
1. Assign default value for optional args explicitly, i.e., ` = None`
2. Update default httpconfig to `configuration = core.HttpConfigurationManager.get_configuration()` to work with SystemLink client and Jupyter Executions.

adamarnesen: Good catch. Updated!

nisystemlink/clients/testmonitor/_test_monitor_client.py
```diff
@@ -23,6 +37,159 @@ def api_info(self) -> models.ApiInfo:
...
+
+        Raises:
+            ApiException: if unable to communicate with the ``/nitestmonitor`` Service or provided
+            invalid arguments.
+        """
+        ...
+
+    @post("update-products", args=[Field("products"), Field("replace")])
+    def update_products(
+        self, products: List[Product], replace: bool = False
```
santhoshramaraj, line 144:
```suggestion
        self, products: List[Product], replace: Optional[bool] = False
```

nisystemlink/clients/testmonitor/_test_monitor_client.py
```diff
@@ -23,6 +37,159 @@ def api_info(self) -> models.ApiInfo:
...
+    ) -> models.CreateProductsPartialSuccess:
+        """Updates a list of products with optional field replacement.
+
+        Args:
+            `products`: A list of products to update. Products are matched for update by id.
+            `replace`: Replace the existing fields instead of merging them.
```
santhoshramaraj, line 150:
```suggestion
            `replace`: Replace the existing fields instead of merging them. Defaults to  `False`.
```

nisystemlink/clients/testmonitor/_test_monitor_client.py
```diff
@@ -23,6 +37,159 @@ def api_info(self) -> models.ApiInfo:
...
+            failures.
+
+        Raises:
+            ApiException: if unable to communicate with the ``/nitestmonitor`` Service
+                or provided an invalid argument.
+        """
+        ...
+
+    @delete("products/{id}")
+    def delete_product(self, id: str) -> Optional[ApiError]:
```
santhoshramaraj, line 166: The success case is `None`, and the failure case is an error, then the Exception occurs.

`ApiError` is never returned by the function call.

```suggestion
    def delete_product(self, id: str) -> None:
```

nisystemlink/clients/testmonitor/_test_monitor_client.py
```diff
@@ -23,6 +37,159 @@ def api_info(self) -> models.ApiInfo:
...
+        Raises:
+            ApiException: if unable to communicate with the ``/nitestmonitor`` Service
+                or provided an invalid argument.
+        """
+        ...
+
+    @post("delete-products", args=[Field("ids")])
+    def delete_products(
+        self, ids: List[str]
+    ) -> Optional[models.DeleteProductsPartialSuccess]:
```
santhoshramaraj, line 181: This function either returns `DeleteProductsPartialSuccess` or results in an Exception, I believe there is no case where it returns `None`.

```suggestion
    ) -> models.DeleteProductsPartialSuccess:
```

nisystemlink/clients/testmonitor/_test_monitor_client.py
```diff
@@ -23,6 +37,159 @@ def api_info(self) -> models.ApiInfo:
...
+        """Deletes multiple products.
+
+        Args:
+            ids (List[str]): List of unique IDs of products.
+
+        Returns:
+            A partial success if any products failed to delete, or None if all
+            products were deleted successfully.
```
santhoshramaraj, line 189:
```suggestion
            A response of product deleted with products failed to delete.
```

nisystemlink/clients/testmonitor/models/_delete_products_partial_success.py
```diff
@@ -0,0 +1,17 @@
+from typing import List, Optional
+
+from nisystemlink.clients.core import ApiError
+from nisystemlink.clients.core._uplink._json_model import JsonModel
+
+
+class DeleteProductsPartialSuccess(JsonModel):
+    """The result of deleting multiple tables when one or more tables could not be deleted."""
```
santhoshramaraj, line 8:
```suggestion
    """The result of deleting multiple products when one or more products could not be deleted."""
```

nisystemlink/clients/testmonitor/models/_query_products_request.py
```diff
@@ -0,0 +1,90 @@
...
+    indicated by non-negative integers that are prefixed with the "at" symbol. Each substitution in
+    the given expression will be replaced by the element at the corresponding index (zero-based) in
+    this list. For example, "@0" in the filter expression will be replaced with the element at the
+    zeroth index of the substitutions list.
+    """
+
+
+class QueryProductsRequest(QueryProductsBase):
+
+    order_by: Optional[ProductField] = None
```
santhoshramaraj, line 52: It is not robust to rely on the camel case convertor of `JsonModel` to convert `order_by` into `orderBy`, instead explicitly define a `Field` alias.

```suggestion
    order_by: Optional[ProductField] = Field(None, alias="orderBy")
```

nisystemlink/clients/testmonitor/_test_monitor_client.py
adamarnesen: changed this documentation

nisystemlink/clients/testmonitor/_test_monitor_client.py
adamarnesen: Good catch. Updated

nisystemlink/clients/testmonitor/_test_monitor_client.py
adamarnesen: I've tried several times to make this work and ran into no end of problems with type casting and subclasses. I think that we don't technically need to mark this as optional. It's not _wrong_ to pass a False from this layer. You're correct that it is _allowed_ not to provide this to the REST layer, but providing False by default is ok. If you can show me how to fix this, we can do it, otherwise, I propose this is ok.

nisystemlink/clients/testmonitor/_test_monitor_client.py
adamarnesen: No, when all the deletes succeed, you get `None` returned and a 204.

nisystemlink/clients/testmonitor/_test_monitor_client.py
adamarnesen: This isn't how this delete API works. If all deletes succeed, you don't get anything back

nisystemlink/clients/testmonitor/models/_delete_products_partial_success.py
adamarnesen: oops, good catch

nisystemlink/clients/testmonitor/models/_query_products_request.py
adamarnesen: Thanks. Updated.

nisystemlink/clients/testmonitor/_test_monitor_client.py
mure: https://github.com/prkumar/uplink/issues/278

I filed an issue for this a couple years ago 🙁

mure (commented):
During the planning for Grafana plugins for Test Monitor, there was a discussion of if we should have one "Test Monitor" plugin or separate plugins for products, test results, etc. The consensus was that Products should be its own plugin, because that's the way our customers understand SystemLink. The fact that it's one microservice is an implementation detail. Should we apply the same logic here and have a `ProductClient`?

nisystemlink/clients/testmonitor/models/_query_products_request.py
```diff
@@ -0,0 +1,91 @@
+from enum import Enum
+from typing import List, Optional
+
+from nisystemlink.clients.core._uplink._json_model import JsonModel
+from pydantic import Field
+
+
+class ProductField(Enum):
```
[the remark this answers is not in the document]
adamarnesen: fixed

nisystemlink/clients/testmonitor/utilities.py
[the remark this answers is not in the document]
adamarnesen: @mure do you have an exmaple of how to do this? In the file client, we have this same pattern. All the examples I see are of classes that are referenced in `__init__.py` files. I don't see any examples of modules directly being re-exported.

nisystemlink/clients/testmonitor/models/_create_products_request.py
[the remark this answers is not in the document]
adamarnesen: good catch. Fixed

adamarnesen:
@mure I like the idea of having a `ProductClient` that isn't obviously part of `/testmonitor` I've refactored this part.

nisystemlink/clients/testmonitor/_test_monitor_client.py
```diff
@@ -26,6 +39,159 @@ def api_info(self) -> models.ApiInfo:
...
+        Raises:
+            ApiException: if unable to communicate with the ``/nitestmonitor`` Service
+                or provided an invalid argument.
+        """
+        ...
+
+    @post("query-products")
+    def query_products(
+        self, query: models.QueryProductsRequest
+    ) -> models.PagedProducts:
```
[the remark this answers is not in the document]
adamarnesen: I'd like to get this in soon so that it's not hanging over me still. What is the timeline for @santhoshramaraj getting that Generator change in?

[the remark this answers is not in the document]
santhoshramaraj: @adamarnesen As this is the first release of `ProductClient` it is good to make the generator concept as the default implementation for paged responses, please refer to #75 for the concept for `query_tables`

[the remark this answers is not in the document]
adamarnesen: @santhoshramaraj @mure it looks like in #75 that the generator pattern is just an addon that _uses_ the Paged based method. Therefore, adding generator will not break this API. I think that I'd prefer to add the generator in a separate PR after we have approved naming conventions in #75. It will make the changes clearer.

santhoshramaraj (approved):

nisystemlink/clients/testmonitor/_test_monitor_client.py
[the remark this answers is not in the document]
santhoshramaraj: @adamarnesen, @mure and I were debating whether to make the generator implementation as the default one instead of an addon, so, `query_tables` could be generator and `_query_tables` or `query_tables_paged` could be the underlying paged implementation.

nisystemlink/clients/testmonitor/utilities.py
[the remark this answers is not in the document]
santhoshramaraj: I believe @mure was suggesting a `../clients/product/utilities/__init__.py` to re-export the functions of `../clients/product/utilities/_get_products_linked_to_file.py`

For end-user, `from nisystemlink.clients.product.utilities import get_products_linked_to_file` remain the same, just that directory structure is consistent.

nisystemlink/clients/product/_product_client.py
```diff
@@ -0,0 +1,182 @@
+"""Implementation of TestMonitor Client"""
```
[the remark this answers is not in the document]
adamarnesen: done

nisystemlink/clients/testmonitor/utilities.py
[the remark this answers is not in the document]
adamarnesen: Ahh, that makes _some_ sense. I've implemented that change

nisystemlink/clients/testmonitor/_test_monitor_client.py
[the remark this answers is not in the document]
adamarnesen: @mure  I need some guidance from you on this one. If we want to implement generator pattern, I'd prefer to change the name of the paged functions in this PR and then do a follow-on with the generator addition. That would keep compatibility and let us move forward on this change.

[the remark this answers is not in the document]
santhoshramaraj: My vote is for `query_products_paged`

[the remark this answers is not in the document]
adamarnesen: I've made this change. Closing this thread.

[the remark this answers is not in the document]
mure: I'm in support of this approach 👍

mure (approved):

nisystemlink/clients/product/_product_client.py
```diff
@@ -0,0 +1,182 @@
...
+            ApiException: if unable to communicate with the ``/nitestmonitor`` service of provided invalid
+                arguments.
+        """
+        ...
+
+    @get(
+        "products",
+        args=[Query("continuationToken"), Query("take"), Query("returnCount")],
+    )
+    def get_products(
```
santhoshramaraj, line 53: @adamarnesen , making this method `get_products_paged` will be appropriate as this could also become a generator.

santhoshramaraj (approved):

PR #76, opened 2024-10-30
Feat: Add minimal client implementation for auth services

Giriharan219:

This PR adds some client methods for auth endpoints and testing for these endpoints.

This Minimal client for auth APIs will enable users to more easily interact with that APIs.

API Link: [Swagger-link](https://dev-api.lifecyclesolutions.ni.com/ninbexecution/swagger/index.html?urls.primaryName=Notebook+Execution+Artifact+Service)

### DIFF
--- a/nisystemlink/clients/auth/_auth_client.py
+++ b/nisystemlink/clients/auth/_auth_client.py
@@ -1,0 +1,35 @@
+"""Implementation of AuthClient."""
+
+from typing import Optional
+
+from nisystemlink.clients import core
+from nisystemlink.clients.core._uplink._base_client import BaseClient
+from nisystemlink.clients.core._uplink._methods import get
+
+from . import models
+
+
+class AuthClient(BaseClient):
+    """A set of methods to access the APIs of SystemLink Auth Client."""
+
+    def __init__(self, configuration: Optional[core.HttpConfiguration] = None):
+        """Initialize an instance.
+
+        Args:
+            configuration: Defines the web server to connect to and information about
+                how to connect. If not provided, an instance of
+                :class:`JupyterHttpConfiguration <nisystemlink.clients.core.JupyterHttpConfiguration>` # noqa: W505
+                is used.
+
+        Raises:
+            ApiException: if unable to communicate with the Auth Service.
+        """
+        if configuration is None:
+            configuration = core.JupyterHttpConfiguration()
+
+        super().__init__(configuration, base_path="/niauth/v1/")
+
+    @get("auth")
+    def authenticate(self) -> models.AuthInfo:
+        """Authenticates the given x-ni-api-key and returns information about the caller."""
+        ...
--- a/nisystemlink/clients/auth/models/_auth_models.py
+++ b/nisystemlink/clients/auth/models/_auth_models.py
@@ -1,0 +1,171 @@
+"""Models utilized for Auth in SystemLink."""
+
+from __future__ import annotations
+
+from datetime import datetime
+from enum import Enum
+from typing import Any, Dict, List, Optional
+
+from nisystemlink.clients.core._uplink._json_model import JsonModel
+from pydantic import Field
+
+
+class AuthStatement(JsonModel):
+    """Auth Statement information."""
+
+    actions: Optional[List[str]] = None
+    """A list of actions the user is allowed to perform."""
+    resource: Optional[List[str]] = None
+    """A list of resources the user is allowed to access."""
+    workspace: Optional[str] = Field(None, example="workspace-id")
+    """The workspace the user is allowed to access."""
+
+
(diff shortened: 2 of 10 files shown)


nisystemlink/clients/auth/models/_auth_models.py
```diff
@@ -0,0 +1,171 @@
...
+
+
+class AuthStatement(JsonModel):
+    """Auth Statement information."""
+
+    actions: Optional[List[str]] = None
+    """A list of actions the user is allowed to perform."""
+    resource: Optional[List[str]] = None
+    """A list of resources the user is allowed to access."""
+    workspace: Optional[str] = Field(None, example="workspace-id")
```
santhoshramaraj, line 20:
```suggestion
    workspace: Optional[str] = None
```
Defining the example in `Field` is not useful as it is not parsed for type hinting.

nisystemlink/clients/auth/models/_auth_models.py
```diff
@@ -0,0 +1,171 @@
...
+    """Statement information."""
+
+    actions: Optional[List[str]] = None
+    """A list of actions the user is allowed to perform."""
+    resource: Optional[List[str]] = None
+    """A list of resources the user is allowed to access."""
+    workspace: Optional[str] = Field(None, example="workspace-id")
+    """The workspace the user is allowed to access."""
+    description: Optional[str] = None
+    """A description for this statement."""
```
santhoshramaraj, line 42: Simplify the definition
```suggestion
class Statement(AuthStatement):
    """Statement information."""

    description: Optional[str] = None
    """A description for this statement."""
```

nisystemlink/clients/auth/models/_auth_models.py
santhoshramaraj, line 1: Remove `Field` definitions if it is used only to store example, instead move the example to docstring to be helpful during type hinting.

nisystemlink/clients/auth/models/_auth_models.py
```diff
@@ -0,0 +1,171 @@
...
+
+
+class Policy(JsonModel):
```
santhoshramaraj, line 45: `Policy` is not referenced, please remove if not used.

santhoshramaraj (changes requested):

ChrisStrykesAgain:
Most of this functionality seems pretty core to the fundamental operations. Would it make more sense to bake this into a base class shared across the various APIs instead of having an explicit API just for this?

ChrisStrykesAgain:
@santhoshramaraj @spanglerco @cameronwaterman I need this functionality for a project I'm working on. If you're OK with the functionality being built into the base class (my preference per my last comment), I'll implement it here pretty quickly. If not, can we get this PR approved and/or any last comments/changes requested? Thanks!

ChrisStrykesAgain:
Actually, thinking through this a little: 

The thing I need, and what I think should really be in the base class is a "get workspace by name" method. The cleanest approach here would probably be to keep auth as it's own class (so basically what's in this PR) and add a class member variable of this type to the base class. I'd prefer to create a wrapper over the "get workspace" function to promote it to a first-class member, but then you'd still have the auth methods available to you as calls on the class member itself. 

Does that seem reasonable?

nisystemlink/clients/auth/_auth_client.py
```diff
@@ -0,0 +1,35 @@
...
+from typing import Optional
+
+from nisystemlink.clients import core
+from nisystemlink.clients.core._uplink._base_client import BaseClient
+from nisystemlink.clients.core._uplink._methods import get
+
+from . import models
+
+
+class AuthClient(BaseClient):
```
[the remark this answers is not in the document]
ChrisStrykesAgain: @santhoshramaraj I think that it's worth leaving this as-is (auth), then creating a secondary API  specifically to expose access control functionality (creating/changing roles and/or workspace permissions management), as they are closely related, but really two different concepts.

As I mentioned in my comments, I think the auth functionality is pretty fundamental and might be worth somehow incorporating into the base class itself, but access control definitely doesn't belong there.

nisystemlink/clients/auth/_auth_client.py
```diff
@@ -0,0 +1,40 @@
...
+    def __init__(self, configuration: Optional[core.HttpConfiguration] = None):
+        """Initialize an instance.
+
+        Args:
+            configuration: Defines the web server to connect to and information about
+                how to connect. If not provided, an instance of
+                :class:`JupyterHttpConfiguration <nisystemlink.clients.core.JupyterHttpConfiguration>` # noqa: W505
```
ChrisStrykesAgain, line 19: I think this would us an instance of `HttpConfiguration` and not `JupyterHttpConfiguration`, right? Or am I missing something?

PR #82, opened 2024-11-07
feat: Add client for SystemLink results API

Madhan-Reddy-ni:

This PR adds client methods for all of the `nitestmonitor` result endpoints as well as testing for these endpoints.

This is an enhancement to the existing test monitor client and will enable users to more easily interact with SystemLink test results.

API Swagger Link - [Results](https://dev-api.lifecyclesolutions.ni.com/niapis/?urls.primaryName=Test+Monitor#/results)


examples/result/results.py
```diff
@@ -0,0 +1,87 @@
...
+# Get all the results using the continuation token in batches of 100 at a time.
+response = client.get_results(take=100, return_count=True)
+all_results = response.results
+while response.continuation_token:
+    response = client.get_results(
+        take=100, continuation_token=response.continuation_token, return_count=True
+    )
+    all_results.extend(response.results)
+
+create_response = create_some_results()
```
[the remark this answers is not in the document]
Madhan-Reddy-ni: Updated to call the create_results first and then the get_results..

santhoshramaraj (approved):

adamarnesen (approved):

nisystemlink/clients/result/models/_query_results_request.py
```diff
@@ -0,0 +1,141 @@
...
+    LEXICOGRAPHIC = "LEXICOGRAPHIC"
+
+
+class QueryResultsBase(JsonModel):
+    filter: Optional[str] = None
+    """
+    The result query filter in Dynamic Linq format.
+    Allowed properties in the filter are:
+    - `id`: String for the global identifier of the result
+    - `status`: String for the status of the result
```
rbell517, line 38: Status is actually a two-field object with a string statusName field and an enum statusType field. The name field is fully settable but should be ignored unless statusType is set to Custom. Generally queries should be checking the statusType field. 

image

You can point them to the `StatusType` model class

nisystemlink/clients/result/models/_query_results_request.py
```diff
@@ -0,0 +1,141 @@
...
+    """
+    descending: Optional[bool] = None
+    """Specifies whether to return the results in descending order.
+    By default, this value is `false` and results are sorted in ascending order.
+    """
+    take: Optional[int] = None
+    """Maximum number of results to return in the current API response.
+    Uses the default if the specified value is negative. The default value is `1000` results.
+    """
+    continuation_token: Optional[str] = None
```
rbell517, line 122: We prefer to base off of `WithPaging` to get the continuation token field. That will help when we have a more reusable paging implementation.

nisystemlink/clients/result/models/_result.py
```diff
@@ -0,0 +1,89 @@
...
+    operator: Optional[str]
+    """The operator that ran the result."""
+
+    file_ids: Optional[List[str]]
+    """A list of file ids that are attached to this result."""
+
+    data_table_ids: Optional[List[str]]
+    """A list of data table ids that are attached to this result."""
+
+    status_type_summary: Optional[Dict[str, int]]
```
rbell517, line 82: A dictionary of string to int is what the swagger types this as, but in practice the key will be the status type enum

nisystemlink/clients/result/models/_result.py
```diff
@@ -0,0 +1,89 @@
...
+    """A summary of the status types in the result."""
+
+    workspace: Optional[str]
+    """The id of the workspace that this product belongs to."""
+
+    is_finalized: Optional[bool]
```
rbell517, line 88: Finalized never made it to the public API documentation. Its a half-baked feature right now, so I would leave it off the client object to avoid any confusion.

nisystemlink/clients/result/_result_client.py
```diff
@@ -0,0 +1,182 @@
...
+
+        Args:
+            query : The query for the fields.
+
+        Returns:
+            A list of the values of the queried field.
+
+        Raises:
+            ApiException: if unable to communicate with the ``/nitestmonitor`` Service or provided
+            invalid arguments.
```
rbell517, line 125:
```suggestion
                invalid arguments.
```

examples/result/results.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+def create_some_results():
+    """Create two example results on your server."""
+    new_results = [
+        Result(
+            part_number="Example 123 AA",
+            program_name=program_name,
+            host_name=host_name,
+            status=StatusObject(status_type=StatusType.PASSED, status_name="Passed"),
```
rbell517, line 23: Can we make this easier? Can this be settable with just the StatusType enum OR the full status object? If we do this we should make it clear in the example, with using just the StatusType enum and one setting a custom status that uses the full object

examples/result/results.py
```diff
@@ -0,0 +1,87 @@
...
+        take=100, continuation_token=response.continuation_token, return_count=True
+    )
+    all_results.extend(response.results)
+
+# use get for first result created
+created_result = client.get_result(create_response.results[0].id)
+
+# Query results without continuation
+query_request = QueryResultsRequest(
+    filter=f'programName="{program_name}" && hostName="{host_name}"',
```
rbell517, line 63: A filter that filters by status would be a good example, since its a bit different and I expect it to be among the more common filter clauses

nisystemlink/clients/result/_result_client.py
```diff
@@ -0,0 +1,182 @@
...
+        """
+        ...
+
+    @post("update-results", args=[Field("results"), Field("replace")])
+    def update_results(
+        self, results: List[Result], replace: bool = False
+    ) -> models.CreateResultsPartialSuccess:
```
rbell517, line 132: Update shouldn't return the create response for updates. I think this is really just point to the name being off. The service uses the same object for both. The swagger has them as different response objects, but I think thats OK to ignore. It properly uses the same type in the results collection in both.

nisystemlink/clients/result/_result_client.py
```diff
@@ -0,0 +1,182 @@
...
+
+        Raises:
+            ApiException: if unable to communicate with the ``/nitestmonitor`` Service or provided
+            invalid arguments.
+        """
+        ...
+
+    @post("update-results", args=[Field("results"), Field("replace")])
+    def update_results(
+        self, results: List[Result], replace: bool = False
```
rbell517, line 131: Technically the allowed values for a result update is not identical to the returned result. You can't directly update updatedAt or the status summary for example. Those are enough of an edge case that I wouldn't want to make this client harder to use to account for those. For example, I wouldn't want to require a separate mapping call to map from the Result returned from a query to the the Result to provide to the update. I like that you can just set the new value you want and send the update.

The other problem with this is the workspace field. The service will change the workspace permission you're required to have based on whether the workspace field is included in the update. If you include workspace, you now need delete and create privileges in that workspace to perform the update, even if the workspace isn't being changed. This is because we don't know whether the workspace is being updated ahead of time. 

This has tripped up a good number of people, particularly if they are operating under the automated agent role that does not have delete privileges. Maybe the right answer is to actually change the service behavior to only require delete/create when the workspace is being changed. That wouldn't be a trivial change.

We have a couple options. For the first issue we can implicitly/silently drop fields from Result that aren't allowed to be updated. That could be confusing; would documentation be good enough? We could raise an error, but now you're back to need another step between query and update. For the second issue, if we're not going to change the service behavior the only improvement I have is to add another boolean allow_workspace_update field to the function params that would keep workspace fields if true, otherwise would replace them with None.

@adamarnesen I'm interested in an outside perspective on this.

nisystemlink/clients/result/models/_query_results_request.py
```diff
@@ -0,0 +1,141 @@
+from enum import Enum
+from typing import List, Optional
+
+from nisystemlink.clients.core._uplink._json_model import JsonModel
+from pydantic import Field
+
+
+class ResultField(str, Enum):
```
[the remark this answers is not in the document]
rbell517: If I had my way I'd say projections should be required to be provided by callers, but its probably a bridge to far.

tests/integration/result/test_result_client.py
```diff
@@ -0,0 +1,368 @@
...
+            ),
+        ]
+        create_results(results)
+
+        get_response = client.get_results_paged(take=1)
+
+        assert get_response is not None
+        assert len(get_response.results) == 1
+
+    def test__create_multiple_results_and_get_results_with_count_at_least_one_count(
```
rbell517, line 149: Make sure you're using the given__when__then naming convention

rbell517 (changes requested):

nisystemlink/clients/result/_result_client.py
priyadarshini-ni: The response object's name for create and update has been changed to `ResultsPartialSuccess` to use for both cases

examples/result/results.py
priyadarshini-ni: The filter has been modified to filter by status. eg.status.statusType == \"Passed\"

examples/result/results.py
priyadarshini-ni: Based on the second suggestion, the examples have been provided to showcase how the `StatusObject` can be set using only the `StatusType` enum and also both `StatusType` and `StatusName` in case of custom value.

nisystemlink/clients/result/models/_query_results_request.py
priyadarshini-ni: The status field has been pointed to the `StatusType` model class

nisystemlink/clients/result/models/_query_results_request.py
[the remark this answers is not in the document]
priyadarshini-ni: The Result Query OrderBy Field enum and Result Values Query Field enum have been made separate to show the distinction.

The ResultField.Product field has also been added to the mentioned enums.

Projection has also been included to the query-results API with the supported values as per swagger.

nisystemlink/clients/result/_result_client.py
priyadarshini-ni: Based on the suggestion for updating workspace, `allow_workspace_update` field has been included to the function params.

For the first issue, would you recommend creating separate request models for UpdateResults and QueryResults to distinguish the fields that can be updated from the query results?

nisystemlink/clients/result/models/_result.py
priyadarshini-ni: This has been removed from the client model

nisystemlink/clients/result/_result_client.py
priyadarshini-ni: For the initial problem mentioned, another model for UpdateRequest has been created to differentiate between the fields that can be updated in a query response. Dropping the fields from Result or adding more documentation could end up being more confusing as you mentioned.

For updating workspace field, `allow_workspace_update` field has been included to the function params which would perform as suggested.

adamarnesen: I think having a separate model is the better approach here. Trying to remove fields from the response model would be confusing. 

Maybe in the future we add a helper that allows you to easily create an update model from a response model and that's how you "downcast" to get rid of the fields. 

Is there an option to have the `update model` be a parent class of the `response model`?

rbell517: It should be doable to make an update request from an existing `Result` but we wouldn't want to make the update model the base class of `Result`, I think that would be confusing.

nisystemlink/clients/result/_result_client.py
```diff
@@ -0,0 +1,214 @@
...
+                before updating.
+
+        Returns: A list of updates results, results that failed to update, and errors for
+            failures.
+
+        Raises:
+            ApiException: if unable to communicate with the ``/nitestmonitor`` Service
+                or provided an invalid argument.
+        """
+        if not allow_workspace_update:
```
rbell517, line 178: Make sure to include test coverage of this change if we keep it around

nisystemlink/clients/result/_result_client.py
```diff
@@ -0,0 +1,214 @@
...
+            The single result matching `id`
+
+        Raises:
+            ApiException: if unable to communicate with the ``/nitestmonitor`` Service
+                or provided an invalid argument.
+        """
+        ...
+
+    @post("query-results")
+    def query_results_paged(
```
rbell517, line 95: I'd rather this just be query_results. That is inconsistent with products but we'll get products changed to match this

rbell517: Same for get_results_paged -> get_results

nisystemlink/clients/result/_result_client.py
rbell517: With the separate update request class I think we can remove the allow_workspace_update parameter from the update_result function, since those update request objects need to be explicitly constructed. When we add back an easy way to create an update request from a `Result` we can revisit this. It will probably make more sense to include that allow_workspace_update field on the function that creates the update request from the Result instead of on the update_result function directly.

rbell517 (changes requested):

nisystemlink/clients/result/models/_update_result_request.py
```diff
@@ -0,0 +1,54 @@
...
+from typing import Dict, List, Optional
+
+from nisystemlink.clients.core._uplink._json_model import JsonModel
+from nisystemlink.clients.result.models._result import StatusObject
+
+
+class UpdateResultRequest(JsonModel):
+    """Contains information about a result."""
+
+    status: StatusObject
```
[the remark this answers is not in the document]
priyadarshini-ni: Updated

nisystemlink/clients/result/models/_result.py
```diff
@@ -0,0 +1,90 @@
...
+    """The type of status."""
+
+    status_name: Optional[str]
+    """The name of the status."""
+
+
+class Result(JsonModel):
+    """Contains information about a result."""
+
+    status: StatusObject
```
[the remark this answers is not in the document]
priyadarshini-ni: All fields have been marked optional

docs/getting_started.rst
```diff
@@ -253,3 +253,29 @@ Delete a feed.
...
    :language: python
    :linenos:
+
+Result API
+-------
+
+Overview
+~~~~~~~~
+
+The :class:`.ResultClient` class is the primary entry point of the Result API.
```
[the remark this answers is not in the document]
priyadarshini-ni: Modified the docs

nisystemlink/clients/result/_result_client.py
```diff
@@ -0,0 +1,214 @@
...
+    Result,
+    UpdateResultRequest,
+)
+from uplink import Field, Query, retry, returns
+
+from . import models
+
+
+@retry(when=retry.when.status([429, 503, 504]), stop=retry.stop.after_attempt(5))
+class ResultClient(BaseClient):
```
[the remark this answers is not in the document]
priyadarshini-ni: The client functions of ResultClient have been moved to TestMonitorClient along with the models

nisystemlink/clients/result/_result_client.py
priyadarshini-ni: The update model and base model have been made into separate models. As update API would be using a separate request model, the `allow_workspace_update` parameter has been removed.

nisystemlink/clients/testmonitor/models/_query_results_request.py
```diff
@@ -0,0 +1,184 @@
...
+
+class ComparisonType(str, Enum):
+    """The valid ways to order a result query."""
+
+    DEFAULT = "DEFAULT"
+    NUMERIC = "NUMERIC"
+    LEXICOGRAPHIC = "LEXICOGRAPHIC"
+
+
+class Projection(str, Enum):
```
SSSantosh18, line 50:
```suggestion
class ResultProjection(str, Enum):
```

nisystemlink/clients/testmonitor/models/_result.py
```diff
@@ -0,0 +1,90 @@
...
+    """A list of data table ids that are attached to this result."""
+
+    status_type_summary: Optional[Dict[StatusType, int]]
+    """A summary of the status types in the result."""
+
+    workspace: Optional[str]
+    """The id of the workspace that this product belongs to."""
+
+    class Config:
+        extra = Extra.allow
```
SSSantosh18, line 90: is this to not face any validation error due to not having an attribute for IsFinalized?

Looks like it will allow any new parameters to be automatically added to the object without throwing validation error. Should we set this to [Ignore](https://docs.pydantic.dev/latest/api/config/#pydantic.config.ConfigDict.extra) so that fields like IsFinalized is not included in the response object?

nisystemlink/clients/testmonitor/models/_results_partial_success.py
```diff
@@ -0,0 +1,20 @@
...
+from nisystemlink.clients.core import ApiError
+from nisystemlink.clients.core._uplink._json_model import JsonModel
+from nisystemlink.clients.testmonitor.models import Result
+
+
+class ResultsPartialSuccess(JsonModel):
+    results: List[Result]
+    """The list of results that were successfully created."""
+
+    failed: Optional[List[Result]] = None
```
SSSantosh18, line 12: failed should point to the request object because it should be same as the request body. There will be additional params in the result object

SSSantosh18: there should be two differnet models for create and update

tests/integration/testmonitor/test_testmonitor.py
```diff
@@ -9,9 +25,378 @@ def client(enterprise_config: HttpConfiguration) -> TestMonitorClient:
...
+            host_name=result.host_name,
+            part_number=result.part_number,
+            serial_number=result.serial_number,
+            total_time_in_seconds=result.total_time_in_seconds,
+            keywords=result.keywords,
+            properties=result.properties,
+            operator=result.operator,
+            file_ids=result.file_ids,
+            data_table_ids=result.data_table_ids,
+        )
```
SSSantosh18, line 402: we can map it with more simple code like

```
UpdateResultRequest(**result.dict())
```

SSSantosh18 (approved):

examples/result/results.py
```diff
@@ -0,0 +1,86 @@
+from nisystemlink.clients.core import HttpConfiguration
```
SSSantosh18, line 1: can we move this under testmonitor folder

santhoshramaraj:
@SSSantosh18 @priyadarshini-ni @rbell517 , when creating the `ProductsClient` it was agreed that it is better to avoid the `TestMonitorClient` concept as it may not be inituitive to users as they don't see the `TestMonitor` anywhere on the webpage and it is best to use separate clients for Products, Results and Steps.

tests/integration/testmonitor/test_testmonitor.py
priyadarshini-ni: Simplified the mapping

nisystemlink/clients/testmonitor/models/_results_partial_success.py
priyadarshini-ni: Created new models for Create and Update partial success reponse. The failed within that model has also been pointed to the request body

nisystemlink/clients/testmonitor/models/_query_results_request.py
priyadarshini-ni: Updated

examples/result/results.py
priyadarshini-ni: The example for `results.py` has been moved under `testmonitor` folder. The file name has been retained.

nisystemlink/clients/testmonitor/models/_result.py
```diff
@@ -0,0 +1,90 @@
...
+
+class StatusObject(JsonModel):
+    """Contains information about a status object."""
+
+    status_type: StatusType
+    """The type of status."""
+
+    status_name: Optional[str]
+    """The name of the status."""
+
```
santhoshramaraj, line 33: Create a constant class for all standard `StatusType` as in most use cases users don't use the `status_name` and it is an additional step to create a `StatusObject`

```suggestion
class StandardResultStatus:
    LOOPING = StatusObject(StatusType.LOOPING)
    SKIPPED = StatusObject(StatusType.SKIPPED)
    ...
   
```

santhoshramaraj (dismissed):

nisystemlink/clients/testmonitor/_test_monitor_client.py
```diff
@@ -40,3 +46,152 @@ def api_info(self) -> models.ApiInfo:
...
+            The single result matching `id`
+
+        Raises:
+            ApiException: if unable to communicate with the ``/nitestmonitor`` Service
+                or provided an invalid argument.
+        """
+        ...
+
+    @post("query-results")
+    def query_results(self, query: models.QueryResultsRequest) -> models.PagedResults:
```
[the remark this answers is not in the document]
priyadarshini-ni: Please refer to this comment where it was suggested to removed `paged` - https://github.com/ni/nisystemlink-clients-python/pull/82#discussion_r1970624403. 
Is there any reasoning for including it?

nisystemlink/clients/testmonitor/models/_result.py
priyadarshini-ni: Added as suggested

nisystemlink/clients/testmonitor/models/_result.py
```diff
@@ -0,0 +1,90 @@
...
+    PASSED = "PASSED"
+    FAILED = "FAILED"
+    RUNNING = "RUNNING"
+    WAITING = "WAITING"
+    TERMINATED = "TERMINATED"
+    ERRORED = "ERRORED"
+    TIMED_OUT = "TIMED_OUT"
+
+
+class StatusObject(JsonModel):
```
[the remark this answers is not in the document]
priyadarshini-ni: Updated

examples/testmonitor/results.py
```diff
@@ -0,0 +1,86 @@
...
+            status=StatusObject(status_type=StatusType.CUSTOM, status_name="Custom"),
+            keywords=["original keyword"],
+            properties={"original property key": "original"},
+        ),
+    ]
+    create_response = client.create_results(new_results)
+    return create_response
+
+
+# Setup the server configuration to point to your instance of SystemLink Enterprise
```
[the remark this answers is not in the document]
priyadarshini-ni: Updated as suggested

priyadarshini-ni:
> @SSSantosh18 @priyadarshini-ni @rbell517, when creating the `ProductsClient`, it was agreed that it is better to avoid the `TestMonitorClient` concept, as it may not be intuitive to users, as they don't see the `TestMonitor` anywhere on the SLE webpage. It is better to use separate clients for Products, Results, and Steps for the user to relate to the different pages or hierarchy.
> 
> FYI - @adamarnesen
Related comment - https://github.com/ni/nisystemlink-clients-python/pull/82#discussion_r1970617384. Confirmation on this would be helpful to proceed further.

nisystemlink/clients/testmonitor/_test_monitor_client.py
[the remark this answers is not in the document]
SSSantosh18: @santhoshramaraj, if not mentioned paged, do you feel a user might confuse this as a client method that returns all items at once?

nisystemlink/clients/result/_result_client.py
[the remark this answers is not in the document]
rbell517: > @SSSantosh18 @priyadarshini-ni @rbell517, when creating the ProductsClient, it was agreed that it is better to avoid the TestMonitorClient concept, as it may not be intuitive to users, as they don't see the TestMonitor anywhere on the SLE webpage. It is better to use separate clients for Products, Results, and Steps for the user to relate to the different pages or hierarchy.
> 
> FYI - @adamarnesen

We don't show Test Monitor in our SLE UI, but it is what the API is named in the Swagger page, which is our official API documentation. Since this a client for that API I think makes sense to name the client accordingly. I also want to reinforce that steps and results are concepts that work together and putting them on the same client does that.

nisystemlink/clients/testmonitor/models/_result.py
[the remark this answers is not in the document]
rbell517: This would mean we'll end up with a separate StepStatus when we add steps. The API uses the same type for both, but them being separate in the client isn't the end of the world. I would good with just calling this `Status` and moving it to its own file. With the other change to add the static helpers to this class you'd get a nice `Status.PASSED` experience, which is great.

rbell517 (approved):

nisystemlink/clients/testmonitor/models/_result.py
priyadarshini-ni: Moved to `Status` class

nisystemlink/clients/testmonitor/_test_monitor_client.py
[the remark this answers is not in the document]
priyadarshini-ni: The `*_paged` nomenclature has been removed from the results client. Similar changes for products API can be taken up in a different PR.

PR #87, opened 2024-12-11
feat: Add client for Notebook & Notebook Execution services.

ganesh-nithin:

TODO: Include high-level description of the changes in this pull request.

TODO: Justify why this contribution should be part of the project.

TODO: Detail what testing has been done to ensure this submission meets requirements.


ChrisStrykesAgain:
In the interest of SLS compatibility, would it make sense to implement this using the webapp service instead of the Notebook service? If you did it in a base class that this inherited from, that could also set the stage for WebVI/SPAs interactions.

adamarnesen:
> In the interest of SLS compatibility, would it make sense to implement this using the webapp service instead of the Notebook service? If you did it in a base class that this inherited from, that could also set the stage for WebVI/SPAs interactions.

I think that this is just trying to add a good wrapper on the HTTP APIs for notebook execution. I think probably the compatibility layer with SLS would be on top of these APIs.

rbell517:
Sorry for the delay on this, I'm getting to this PR now. 

@adamarnesen @ChrisStrykesAgain on the topic of SLS compatibility -- 
We do not have API compatibility between SLE and SLS notebook APIs, both the notebook resource APIs and the notebook exec APIs. They have totally different implementations between the two and different API capabilities. We could theoretically make an adapter layer that adapts the client to call the different APIs with the different models, but there are capability gaps between the two that we couldn't overcome. I will not hold up getting SLE support included in this library to include this kind of adapter.

As for exposing the webapp API for accessing notebooks instead for SLE, I'd rather we not do that. Yes, do we store notebooks in the webapp service, but that shouldn't be assumed by the client. We intentionally have made changes to the SLE notebooks API from the webapp service API and the client should continue with that direction. I think we will want an entirely different client for webapps.

rbell517 (changes requested):

rbell517:
This has been superseded by #105 so I'm closing this PR.

PR #89, opened 2025-01-08
feat: Add Client for Systems State Service APIs

shri2k2:

This PR adds API Client Module for `States` microservice in the `Systems State Service`.

The `SystemsStateClient` makes it easier for users to interact with the endpoints of `Systems State Service`  by just creating an instance of it.

API Link: [Swagger Link](https://dev-api.lifecyclesolutions.ni.com/niapis/?urls.primaryName=Systems+State+Service)

### DIFF
--- a/nisystemlink/clients/systemsstate/_systems_state_client.py
+++ b/nisystemlink/clients/systemsstate/_systems_state_client.py
@@ -1,18 +1,18 @@
 """Implementation of SystemsStateService Client"""
 
-from typing import Dict, List, Optional
+from io import BytesIO
+from typing import Any, Dict, List, Optional
 
 from nisystemlink.clients import core
 from nisystemlink.clients.core._uplink._base_client import BaseClient
 from nisystemlink.clients.core._uplink._methods import delete, get, patch, post
-from uplink import Body, Part, Path, Query
+from uplink import Body, multipart, Part, Path, Query
+
 
 from . import models
 
 
 class SystemsStateClient(BaseClient):
-    # prevent pytest from thinking this is a test class
-    __test__ = False
 
     def __init__(self, configuration: Optional[core.HttpConfiguration] = None):
         """Initialize an instance.
@@ -31,16 +31,6 @@ class SystemsStateClient(BaseClient):
 
         super().__init__(configuration, base_path="/nisystemsstate/v1/")
 
-    @get(
-        "states",
-        args=[
-            Query("Skip"),
-            Query("Take"),
-            Query("Workspace"),
-            Query("Architecture"),
-            Query("Distribution"),
-        ],
-    )
     def get_states(
         self,
         skip: Optional[int] = None,
@@ -65,10 +55,44 @@ class SystemsStateClient(BaseClient):
             ApiException : if unable to communicate with the ``/nisystemsstate`` Service
                 or provided an invalid argument.
         """
-        ...
+        """Uplink does not support enum serializing into str
+
+        The below inner function is the uplink executable and the outer one acts as a wrapper function,
+        which is exposed to the end user
+        """
+
+        @get(
+            "states",
+            args=[
+                Query("Skip"),
+                Query("Take"),
+                Query("Workspace"),
+                Query("Architecture"),
+                Query("Distribution"),
+            ],
+        )
+        def __get_states(
+            self,
+            skip: Optional[int] = None,
+            take: Optional[int] = None,
+            workspace: Optional[str] = None,
+            architecture: Optional[str] = None,
+            distribution: Optional[str] = None,
+        ) -> models.StateDescriptionListResponse:
+            """Uplink implementation"""
+            ...
+
+        return __get_states(
+            self,
+            skip=skip,
+            take=take,
+            workspace=workspace,
+            architecture=architecture.value if architecture is not None else None,
+            distribution=distribution.value if distribution is not None else None,
+        )
 
     @post("states", args=[Body])
-    def create_state(self, new_state: models.StateRequest) -> models.StateResponse:
+    def create_state(self, state: models.StateRequest) -> models.StateResponse:
         """Create state
 
         Args:
@@ -84,11 +108,11 @@ class SystemsStateClient(BaseClient):
         ...
 
     @post("export-state", args=[Body])
-    def export_state(self, export_state_request: models.ExportStateRequest) -> str:
+    def export_state(self, query: models.ExportStateRequest) -> str:
         """Generate state export
 
         Args:
-            `export_state_request`: Contains identifying information on the state to export.
+            `query`: Contains identifying information on the state to export.
 
         Returns:
             The exported state as a file.
@@ -118,11 +142,11 @@ class SystemsStateClient(BaseClient):
         ...
 
     @get("states/{stateId}", args=[Path("stateId")])
-    def get_state(self, state_id: str) -> models.StateResponse:
+    def get_state(self, id: str) -> models.StateResponse:
         """Get state based on the give stateId parameter
 
         Args:
-            `state_id`: The respective id of the state
+            `id`: The respective id of the state
 
         Returns:
             The state of the given id
@@ -135,12 +159,12 @@ class SystemsStateClient(BaseClient):
 
     @patch("states/{stateId}", args=[Path("stateId"), Body])
     def update_state(
-        self, state_id: str, patch_updates: Dict[str, str]
+        self, id: str, patch_updates: Dict[str, Any]
     ) -> models.StateResponse:
         """Update an existing state
 
         Args:
-            `state_id`: The respective id of the state to be updated
+            `id`: The respective id of the state to be updated
             `patch_updates`: A dictionary containing the properties to update, where
             keys are property names (strings) and values are the updated values.
 
@@ -154,11 +178,11 @@ class SystemsStateClient(BaseClient):
         ...
 
     @delete("states/{stateId}", args=[Path("stateId")])
-    def delete_state(self, state_id: str) -> None:
+    def delete_state(self, id: str) -> None:
         """Deleting an existing state
 
         Args:
-            `state_id`: The respective id of the state to be deleted
+            `id`: The respective id of the state to be deleted
 
         Returns:
             None
@@ -174,11 +198,11 @@ class SystemsStateClient(BaseClient):
         args=[
             Part(name="Id", type=str),
             Part(name="ChangeDescription", type=str),
-            Part(name="File", type=str),
+            Part(name="File", type=BytesIO),
         ],
     )
     def replace_state_content(
-        self, id: str, change_description: str, file: str
+        self, id: str, change_description: str, file: BytesIO
     ) -> models.StateResponse:
         """Replace state content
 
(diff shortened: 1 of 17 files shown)


nisystemlink/clients/systemsstate/_systems_state_client.py
```diff
@@ -0,0 +1,256 @@
...
+        Raises:
+            ApiException : if unable to communicate with the ``/nisystemsstate`` Service
+                or provided an invalid argument.
+        """
+        ...
+
+    @post("export-state-from-system", args=[Body])
+    def export_state_from_system(
+        self, export_state_from_system_request: models.ExportStateFromSystemRequest
+    ) -> str:
```
santhoshramaraj, line 105:
```suggestion
    @params({"inline": True})  # type: ignore
    @post("export-state-from-system", args=[Field("systemID")])
    def export_state_from_system(self, system_id: str) -> str:
```
Eliminate the need for a request object by inlining parameters,
References for the recommendation,
https://github.com/ni/nisystemlink-clients-python/blob/caddc7501b240b8f4653447b7851f40f3c21de78/nisystemlink/clients/file/_file_client.py#L173
https://github.com/ni/nisystemlink-clients-python/blob/caddc7501b240b8f4653447b7851f40f3c21de78/nisystemlink/clients/product/_product_client.py#L31

nisystemlink/clients/systemsstate/_systems_state_client.py
```diff
@@ -0,0 +1,256 @@
...
+            The exported state of the system as a file.
+
+        Raises:
+            ApiException : if unable to communicate with the ``/nisystemsstate`` Service
+                or provided an invalid argument.
+        """
+        ...
+
+    @get("states/{stateId}", args=[Path("stateId")])
+    def get_state(self, state_id: str) -> models.StateResponse:
```
santhoshramaraj, line 121: The method name and arg `id` together implies `state_id`
```suggestion
    def get_state(self, id: str) -> models.StateResponse:
```

nisystemlink/clients/systemsstate/_systems_state_client.py
```diff
@@ -0,0 +1,256 @@
...
+
+        Raises:
+            ApiException : if unable to communicate with the ``/nisystemsstate`` Service
+                or provided an invalid argument.
+        """
+        ...
+
+    @patch("states/{stateId}", args=[Path("stateId"), Body])
+    def update_state(
+        self, state_id: str, patch_updates: Dict[str, str]
```
santhoshramaraj, line 138:
```suggestion
        self, id: str, patch_updates: Dict[str, str]
```

nisystemlink/clients/systemsstate/_systems_state_client.py
```diff
@@ -0,0 +1,256 @@
...
+            The updated state
+
+        Raises:
+            ApiException : if unable to communicate with the ``/nisystemsstate`` Service
+                or provided an invalid argument.
+        """
+        ...
+
+    @delete("states/{stateId}", args=[Path("stateId")])
+    def delete_state(self, state_id: str) -> None:
```
santhoshramaraj, line 157:
```suggestion
    def delete_state(self, id: str) -> None:
```

nisystemlink/clients/systemsstate/_systems_state_client.py
```diff
@@ -0,0 +1,256 @@
...
+
+    @post(
+        "replace-state-content",
+        args=[
+            Part(name="Id", type=str),
+            Part(name="ChangeDescription", type=str),
+            Part(name="File", type=str),
+        ],
+    )
+    def replace_state_content(
```
santhoshramaraj, line 180: By file, do you mean the str content of the file? in the description it states as .sls file, then, should the file be a path?

nisystemlink/clients/systemsstate/_systems_state_client.py
```diff
@@ -0,0 +1,256 @@
...
+            Part(name="Name", type=str),
+            Part(name="Description", type=str),
+            Part(name="Distribution", type=str),
+            Part(name="Architecture", type=str),
+            Part(name="Properties", type=str),
+            Part(name="Workspace", type=str),
+            Part(name="File", type=str),
+        ],
+    )
+    def import_state(
```
santhoshramaraj, line 211: Since there are `models.Architecture` and `models.Distribution`, use the same as inputs for this method as well, do handle the enum to str conversion using the wrapper function

nisystemlink/clients/systemsstate/_systems_state_client.py
```diff
@@ -0,0 +1,256 @@
...
+            The imported state.
+
+        Raises:
+            ApiException : if unable to communicate with the ``/nisystemsstate`` Service
+                or provided an invalid argument.
+        """
+        ...
+
+    @post("delete-states", args=[Body])
+    def delete_states(self, states_id_list: List[str]) -> None:
```
santhoshramaraj, line 243:
```suggestion
    def delete_states(self, ids: List[str]) -> None:
```

nisystemlink/clients/systemsstate/models/_state.py
```diff
@@ -0,0 +1,60 @@
...
+    """
+
+
+class StateMetaData(State):
+    """Model for state metadata information."""
+
+    id: Optional[str] = None
+    """Gets or sets the ID of the state."""
+
+    createdTimestamp: datetime
```
santhoshramaraj, line 43: For all model fields, use snake_case and let the `JsonModel` convert it to camelCase at runtime serialization.

```suggestion
    created_timestamp: datetime
```

camelCase serialization defined at https://github.com/ni/nisystemlink-clients-python/blob/caddc7501b240b8f4653447b7851f40f3c21de78/nisystemlink/clients/core/_uplink/_json_model.py#L4

nisystemlink/clients/systemsstate/models/_state.py
```diff
@@ -0,0 +1,60 @@
...
+    """Gets or sets ISO-8601 formatted timestamp specifying the last date that the state was updated."""
+
+
+class AdditionalStateInformation(JsonModel):
+    """Encompasses the feeds, systemImage and packages associated with a particular state"""
+
+    feeds: Optional[List[Feed]] = None
+    """List of all feeds associated with the particular state"""
+
+    systemImage: SystemImage
```
santhoshramaraj, line 56: make Pythonic
```suggestion
    system_image: SystemImage
```

tests/integration/systemsstate/test_systemsstate.py
```diff
@@ -0,0 +1,242 @@
...
+@pytest.fixture(scope="class")
+def client(enterprise_config: HttpConfiguration) -> SystemsStateClient:
+    """Fixture to create a SystemsStateClient instance."""
+    return SystemsStateClient(enterprise_config)
+
+
+@pytest.fixture
+def default_work_space() -> str:
+    """Fixture that returns the default workspace id"""
+    return "846e294a-a007-47ac-9fc2-fac07eab240e"
```
santhoshramaraj, line 18: Tests should be agnostic of the SLE instance and equipped to work with a fresh deployment by taking care of creating/deleting any necessary entities or mock them in the requests response.
```suggestion
    return ""
```

santhoshramaraj (changes requested):
Pending tests review

shri2k2:
@santhoshramaraj 

I'm yet to update the changes that I made for resolving the previous comments. Awaiting your reply on the privilege issue (test tier for my role) for creating or updating states, to test the modified client code. However, I have few doubts.

shri2k2:
@santhoshramaraj 

In the functions involving `multipart-form-data` as expected request body, in the uplink documentation, it is suggested to use the `@multipart` decorator. But while referring the `feeds` client, I couldn't find such a decorator. Can we avoid that?

If I make use of the decorator, the response is not inferring the expected type properly. Also, the function signature in the documentation is displayed as `(method) import_state: multipart`, instead of showing the proper docstring.

nisystemlink/clients/systemsstate/_systems_state_client.py
shri2k2: `string($binary)` This is the type given in the swagger docs. So, I'm expecting the content of the file itself from the end user, in the format of `io.BytesIO`.

nisystemlink/clients/systemsstate/_systems_state_client.py
```diff
@@ -0,0 +1,256 @@
...
+            Query("Distribution"),
+        ],
+    )
+    def get_states(
+        self,
+        skip: Optional[int] = None,
+        take: Optional[int] = None,
+        workspace: Optional[str] = None,
+        architecture: Optional[models.Architecture] = None,
+        distribution: Optional[models.Distribution] = None,
```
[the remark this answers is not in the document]
shri2k2: In the given reference, I could see that a user face function is defined, which expects the enum. Then it is converted to str and then passed to the uplink-executable function, which is defined as a private function of the class. Also could be able to see similar pattern in other clients too.

Even though, as a consumer of the package we shouldn't make use of the private functions, it'll still be visible right?! Wouldn't it create any visible confusions to the end user ?!

```py
def get_states(
        self,
        skip: Optional[int] = None,
        take: Optional[int] = None,
        workspace: Optional[str] = None,
        architecture: Optional[models.Architecture] = None,
        distribution: Optional[models.Distribution] = None,
    ) -> models.StateDescriptionListResponse:
        """Get the list of states

        Args:
            `skip`: Number of files to skip in the result when paging
            `take`: Number of files to return in the result
            `workspace`: The workspace id of the states
            `architecture`: The states that are compatible with the requested architecture
            `distribution`: The states that are compatible with the requested distribution

        Returns:
            The list of states for the given query parameters

        Raises:
            ApiException : if unable to communicate with the ``/nisystemsstate`` Service
                or provided an invalid argument.
        """

        """Uplink does not support enum serializing into str

        The below inner function is the uplink executable and the outer one acts as a wrapper function,
        which is exposed to the end user
        """

        @get(
            "states",
            args=[
                Query("Skip"),
                Query("Take"),
                Query("Workspace"),
                Query("Architecture"),
                Query("Distribution"),
            ],
        )
        def __get_states(
            self,
            skip: Optional[int] = None,
            take: Optional[int] = None,
            workspace: Optional[str] = None,
            architecture: Optional[str] = None,
            distribution: Optional[str] = None,
        ) -> models.StateDescriptionListResponse:
            ...
        
        return __get_states(
            skip=skip,
            take=take,
            workspace=workspace,
            architecture=architecture.value if architecture is not None else None,
            distribution=distribution.value if distribution is not None else None
        )
      ```
      
Here the uplink-executable isn't a direct private member of the class, rather it is just kind of like an inner helper function. What is your take on this or should we stick to the existing pattern itself.

**Comment by shri2k2** [comment]
@santhoshramaraj 

`stateID: Optional[str] = None` How do the snake case and camel case conversion would be happening for this field? Consider we are defining it as `state_id`, then the camel case conversion would be `stateId` right! Just a small confusion that, wouldn't there be any problem with this! Because, in the swagger docs it is expected to be `stateID`.

**Comment by santhoshramaraj** [comment]
> @santhoshramaraj
> 
> In the functions involving `multipart-form-data` as expected request body, in the uplink documentation, it is suggested to use the `@multipart` decorator. But while referring the `feeds` client, I couldn't find such a decorator. Can we avoid that?
> 
> If I make use of the decorator, the response is not inferring the expected type properly. Also, the function signature in the documentation is displayed as `(method) import_state: multipart`, instead of showing the proper docstring.

Please refer to `upload_file` method at https://github.com/ni/nisystemlink-clients-python/blob/68a42a257244a884ef36e98c0eb6506e282be33f/nisystemlink/clients/file/_file_client.py#L191 to use `multi-part` with Uplink.

File: nisystemlink/clients/systemsstate/_systems_state_client.py (hunk 11 above)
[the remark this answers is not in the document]
Comment (santhoshramaraj) [reply nisystemlink/clients/systemsstate/_systems_state_client.py:50]: As long as this works, I am good with this implementation as `__get_states()` is not used elsewhere in the class.

File: nisystemlink/clients/systemsstate/models/_state_request.py [hunk 12]
```diff
@@ -0,0 +1,46 @@
...
+
+
+class StateRequest(State, AdditionalStateInformation):
+    """Model for system state object."""
+
+
+class ExportStateRequest(JsonModel):
+    """Contains identifying information of the state to export."""
+
+    inline: Optional[bool] = None
```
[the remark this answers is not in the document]
SSSantosh18: @rbell517 were intending to get thoughts from @santhoshramaraj?

PR #96, opened 2025-02-27
feat: Add steps APIs to the testmonitor client

SSSantosh18:

Include methods for `nitestmonitor/v2/steps/` endpoints in testmonitor client

Steps APIs would be frequently used to extract measurement data from SystemLink for usage in custom scripts or notebook. So a Py client for that would make the script developer's life easy

Manual tested using a Py script


nisystemlink/clients/testmonitor/_test_monitor_client.py
```diff
@@ -40,3 +42,164 @@ def api_info(self) -> models.ApiInfo:
...
+        """
+        ...
+
+    @post("update-steps")
+    def update_steps(
+        self, steps: models.UpdateStepsRequest
+    ) -> Optional[models.UpdateStepsPartialSuccess]:
+        """Updates one or more steps.
+
+        Update requires that the version field matches the version being updated from.
```
SSadaiyappan-NI, line 105: Does steps have a version field?

nisystemlink/clients/testmonitor/models/_create_steps_partial_success.py
```diff
@@ -0,0 +1,25 @@
...
+from nisystemlink.clients.core import ApiError
+from nisystemlink.clients.core._uplink._json_model import JsonModel
+from nisystemlink.clients.testmonitor.models._create_steps_request import (
+    CreateStepRequestObject,
+)
+from nisystemlink.clients.testmonitor.models._step import Step
+
+
+class CreateStepsPartialSuccess(JsonModel):
+    steps: Optional[List[Step]] = None
```
SSadaiyappan-NI, line 12: This is not an optional field

nisystemlink/clients/testmonitor/models/_delete_steps_partial_success.py
```diff
@@ -0,0 +1,25 @@
...
+    """The ID of the step."""
+
+    result_id: str
+    """The ID of the result associated with the step."""
+
+
+class DeleteStepsPartialSuccess(JsonModel):
+    """The result of deleting multiple steps when one or more steps could not be deleted."""
+
+    steps: Optional[List[StepIdResultIdPair]] = None
```
SSadaiyappan-NI, line 18: In swagger this is not given as optional, if there are no steps to delete, will it provide the key and empty list, or do we need to keep it optional?

nisystemlink/clients/testmonitor/models/_update_steps_partial_success.py
```diff
@@ -0,0 +1,25 @@
...
+from nisystemlink.clients.core import ApiError
+from nisystemlink.clients.core._uplink._json_model import JsonModel
+from nisystemlink.clients.testmonitor.models._step import Step
+from nisystemlink.clients.testmonitor.models._update_steps_request import (
+    UpdateStepRequestObject,
+)
+
+
+class UpdateStepsPartialSuccess(JsonModel):
+    steps: Optional[List[Step]] = None
```
SSadaiyappan-NI, line 12: same here

nisystemlink/clients/testmonitor/models/_update_steps_request.py
```diff
@@ -0,0 +1,28 @@
...
+    StepRequestObjectBase,
+)
+
+
+class UpdateStepRequestObject(StepRequestObjectBase):
+    name: Optional[str] = None
+    """Step name."""
+
+    children: Optional[List["UpdateStepRequestObject"]] = None
+    """Nested child steps."""
```
SSadaiyappan-NI, line 14: Create step request and update step request are both the same, do we need seperate models for this, else we can use the same

SSadaiyappan-NI (changes requested):

nisystemlink/clients/testmonitor/_test_monitor_client.py
SSSantosh18: No, it doesn't

SSSantosh18:
> We have to include the api reference updates and example updates apart from the tests. If that is due to the dependency with the results PR, I will wait for the update for the next review

1. Added example
2. Added integration tests, Ideally result has to be created and used for step APIs, for now I have hardcoded the result ID. Will update this once Results API is merged
3. Yet to api reference updates

@SSadaiyappan-NI

nisystemlink/clients/testmonitor/models/_delete_steps_partial_success.py
SSSantosh18: it will be empty. Made required

nisystemlink/clients/testmonitor/models/_create_steps_partial_success.py
SSSantosh18: it will be empty. made required

tests/integration/testmonitor/test_testmonitor.py
```diff
@@ -9,9 +32,344 @@ def client(enterprise_config: HttpConfiguration) -> TestMonitorClient:
...
+        self, client: TestMonitorClient, create_steps, unique_identifier
+    ):
+        step_id = unique_identifier
+        result_id = "ea6f8d2c-8d57-441e-8375-aa897f59835e"
+        name = "Test Step 1"
+        data = StepDataObject(
+            text="This is a test step", parameters=[{"name": "param1", "value": "10"}]
+        )
+        properties = {"property1": "value1", "property2": "value2"}
+        # createResultResponse: CreateResultsPartialSuccess = create_result()
```
SSSantosh18, line 118: eventually this create_result() fixture will be used to create result in all tests

examples/testmonitor/steps.py
```diff
@@ -0,0 +1,93 @@
...
+delete_response = client.delete_steps(
+    steps=[
+        StepIdResultIdPair(step_id=step.step_id, result_id=step.result_id)
+        for step in queried_steps
+    ]
+)
+
+create_response = client.create_steps(CreateStepsRequest(steps=step_requests))
+# delete steps one by one
+created_steps = create_response.steps
```
SSadaiyappan-NI, line 90:
```suggestion
created_steps = create_response.steps
# delete steps one by one
```

SSadaiyappan-NI (changes requested):
Will wait for other changes in the PR

tests/integration/testmonitor/test_testmonitor.py
priyadarshini-ni: The `create_results` fixture itself has been used as the functionality for both were same

nisystemlink/clients/testmonitor/models/_create_steps_partial_success.py
```diff
@@ -0,0 +1,25 @@
...
+)
+from nisystemlink.clients.testmonitor.models._step import Step
+
+
+class CreateStepsPartialSuccess(JsonModel):
+    steps: List[Step]
+    """The list of steps that were successfully created."""
+
+    failed: Optional[List[CreateStepRequest]] = None
+    """The list of steps that were not created.
```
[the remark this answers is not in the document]
priyadarshini-ni: Modified to use `list` in all places

SSadaiyappan-NI (approved):

nisystemlink/clients/testmonitor/models/_result.py
```diff
@@ -42,7 +42,7 @@ class Result(JsonModel):
     keywords: Optional[List[str]]
     """A list of keywords that categorize this result."""
 
-    properties: Optional[Dict[str, str]]
+    properties: Optional[Dict[str, Optional[str]]]
```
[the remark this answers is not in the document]
priyadarshini-ni: The change for this has been done in this [PR](https://github.com/ni/nisystemlink-clients-python/pull/99). This change has been added here for the pipeline check

tests/integration/testmonitor/test_testmonitor.py
```diff
@@ -396,3 +436,362 @@ def __map_result_to_update_result_request(
...
+        ]
+        created_result = create_results(results)
+        step_id = unique_identifier
+        result_id = created_result.results[0].id
+        name = "Test Step 1"
+        data = StepDataObject(
+            text="This is a test step", parameters=[{"name": "param1", "value": "10"}]
+        )
+        properties = {"property1": "value1", "property2": "value2"}
+        step = CreateStepRequest(
```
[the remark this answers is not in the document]
priyadarshini-ni: Added

examples/testmonitor/steps.py
```diff
@@ -0,0 +1,94 @@
...
+
+# Setup the server configuration to point to your instance of SystemLink Enterprise
+server_configuration = HttpConfiguration(
+    server_uri="https://yourserver.yourcompany.com",
+    api_key="YourAPIKeyGeneratedFromSystemLink",
+)
+client = TestMonitorClient(configuration=server_configuration)
+
+# Create the step requests
+result_id = "4bef8c3c-afbf-400a-9d30-065dfb0b8d24"
```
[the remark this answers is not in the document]
priyadarshini-ni: Added an example where the result is created and its Id is used to create steps

nisystemlink/clients/testmonitor/models/_step.py
```diff
@@ -0,0 +1,80 @@
...
+
+    value: Any
+    """The value."""
+
+
+class StepData(JsonModel):
+    text: Optional[str] = None
+    """Text string describing the output data."""
+
+    parameters: Optional[List[dict[str, Optional[str]]]] = None
```
rbell517, line 20: I'd like to do better with the steps data than the anything goes the API allows.  We have a standard set of fields to represent certain common parts of a measurement that would go here and the client should encourage using those. What I recommend is a new `Measurement` class with all the standard fields along with an `extras` dictionary for other stuff. 

Here is what I would say is the ideal form.
```python
class Measurement(BaseModel):
    name: Optional[str] = None
    status: Optional[Status] = None
    measurement: Optional[Union[str, int, float]] = None
    lowLimit: Optional[Union[str, int, float]] = None
    highLimit: Optional[Union[str, int, float]] = None
    units: Optional[str] = None
    comparisonType: Optional[str] = None
    extra: Dict[str, str] = {}
```

I say this is the ideal because I'm willing to compromise on this if it makes sense to, for example typing each field as string like it is on the API. I'm concerned about the performance cost for converting these to numeric types if the caller is pulling down a large number of steps with a large number of measurements. Keeping everything as strings would eliminate that. I also don't expect uplink to handle these unions well since the JSON is always going to be a string so it won't know to convert to a number, and neither would we, frankly, we'd basically be guessing. 

cc @adamarnesen and @SSSantosh18 in case they want to add anything. In particular to know how string or numeric typing would impact the data frame we ultimately want to make.

tests/integration/testmonitor/test_testmonitor.py
```diff
@@ -396,3 +439,417 @@ def __map_result_to_update_result_request(
...
+        assert create_response is not None
+        assert len(create_response.steps) == 1
+        step = create_response.steps[0]
+        new_name = "Updated Name"
+
+        update_response: UpdateStepsPartialSuccess = client.update_steps(
+            UpdateMultipleStepsRequest(
+                steps=[
+                    UpdateStepRequest(
+                        step_id=step.step_id, result_id=step.result_id, name=new_name
```
rbell517, line 723: Same as in the example, it would be better to update data and inputs/outputs. That would also verify we are transfering those more complex fields properly.

rbell517 (changes requested):

tests/integration/testmonitor/test_testmonitor.py
SSadaiyappan-NI: updated as per suggestion

nisystemlink/clients/testmonitor/models/_step.py
```diff
@@ -0,0 +1,80 @@
...
+
+class NamedValue(JsonModel):
+    name: str
+    """The name of the value."""
+
+    value: Any
+    """The value."""
+
+
+class StepData(JsonModel):
```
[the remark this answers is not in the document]
SSadaiyappan-NI: created two seperate files for these models

examples/testmonitor/steps.py
```diff
@@ -0,0 +1,115 @@
...
+)
+
+# update the name of a step
+update_response = client.update_steps(
+    UpdateMultipleStepsRequest(
+        steps=[
+            UpdateStepRequest(
+                step_id=step.step_id,
+                result_id=step.result_id,
+                name="updated name",
```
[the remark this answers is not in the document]
SSadaiyappan-NI: updated the data instead of the step name

nisystemlink/clients/testmonitor/_test_monitor_client.py
```diff
@@ -197,3 +197,165 @@ def delete_results(
...
+            ApiException: if unable to communicate with the `/nitestmonitor` service or if there are
+            invalid arguments.
+        """
+        ...
+
+    @delete(
+        "results/{resultId}/steps/{stepId}",
+        args=[Path("resultId"), Path("stepId"), Query("updateResultTotalTime")],
+    )
+    def delete_step(
```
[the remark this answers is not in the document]
SSadaiyappan-NI: updated the order

nisystemlink/clients/testmonitor/models/_step.py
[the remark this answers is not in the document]
SSadaiyappan-NI: Updated

nisystemlink/clients/testmonitor/models/_query_steps_request.py
```diff
@@ -0,0 +1,141 @@
...
+    unspecified fields will be excluded. If no projection is specified,
+    all step fields will be returned.
+    """
+
+    response_format: Optional[StepResponseFormat] = None
+    """Enum indicating the expected response format (CSV or JSON)."""
+
+
+class QueryStepValuesRequest(QueryStepsBase):
+    field: Optional[StepField] = None
```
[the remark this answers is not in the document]
SSadaiyappan-NI: updated it to be required

nisystemlink/clients/testmonitor/models/_query_steps_request.py
```diff
@@ -0,0 +1,141 @@
...
+    order_by: Optional[StepOrderBy] = None
+    """Specifies the fields to use to sort the steps."""
+
+    descending: Optional[bool] = None
+    """Specifies whether to return the steps in descending order."""
+
+    take: Optional[int] = None
+    """Maximum number of steps to return in the current API response."""
+
+    continuation_token: Optional[str] = None
```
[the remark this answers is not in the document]
SSadaiyappan-NI: updated

nisystemlink/clients/testmonitor/models/_query_steps_request.py
```diff
@@ -0,0 +1,141 @@
...
+    WORKSPACE = "WORKSPACE"
+    KEYWORDS = "KEYWORDS"
+    PROPERTIES = "PROPERTIES"
+
+
+class StepResponseFormat(str, Enum):
+    """An enumeration of response formats for step queries."""
+
+    JSON = "JSON"
+    CSV = "CSV"
```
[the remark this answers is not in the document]
SSadaiyappan-NI: removed the response format

nisystemlink/clients/testmonitor/models/_create_steps_request.py
```diff
@@ -0,0 +1,66 @@
...
+
+class CreateStepRequest(BaseStepRequest):
+    name: str
+    """Step name."""
+
+    children: Optional[List["CreateStepRequest"]] = None
+    """Nested child steps."""
+
+
+class CreateMultipleStepsRequest(JsonModel):
```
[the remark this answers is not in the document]
SSadaiyappan-NI: updated

nisystemlink/clients/testmonitor/models/_step.py
SSadaiyappan-NI: image
Hi @rbell517 ,

We have added a config to allow extra fields to be added to the model apart from the ones already mentioned. But one drawback we see here is, we can't validate the types of the extra fields in the current pydantic version (1.10.2) we are using. If we have it as a dictionary like you mentioned, we may have to flatten it before sending it to create steps request and put the extra fields inside the extra after querying the steps. We feel that it may take a hit in the performance when we access the extra fields to flatten it and append it with existing data, if there are lot of measurements with extra fields.

If we just provide `Extra: allow`, we can allow extra fields. 
1. If we need to validate fields, we may need to manually iterate through data of each measurement which might affect performance.
2. One more option is to check out the latest Pydantic version and its impact on upgrading it. It might allow us to set the types for extra fields.

Could you provide your thoughts on this?
FYI - @SSSantosh18

rbell517: > One more option is to check out the latest Pydantic version and its impact on upgrading it.

Unfortunately upgrading to v2 of pydantic has a number of breaking changes so we need to consider the impact.

> they might be able to automatically identify the datatype and consider them as int, float. 

From googling it looks like pandas ability to infer types is so-so, but casting could be reasonable for someone who knows all their measurement data in this analysis workflow are numbers.

I'm OK with keeping the values as strings. I took a look at what pydantic will do with Unions and I agree it will likely be too slow.

For the extras, we still need to be able to access the extra fields for when the Measurement comes as a response. Again newer pydantic versions would help here, they add a native `model_extra` to BaseModel that would surface the extras, but we don't have that. Can you tell me whether extra fields are accessible with what you've implemented? e.g. if the measurement had a `spec_id` field could you access it as `measurement.spec_id`? If so I think that is acceptable.

nisystemlink/clients/testmonitor/models/_step_data.py
```diff
@@ -0,0 +1,32 @@
...
+    units: Optional[str] = None
+    comparisonType: Optional[str] = None
+
+    class Config:
+        extra = Extra.allow
+
+    def __init__(self, **data: Any) -> None:
+        # Convert all extra fields to str while keeping known fields unchanged
+        processed_data = {
+            k: str(v) if k not in self.__annotations__ else v for k, v in data.items()
```
rbell517, line 22: I think you want self.__fields__ instead of __annotations

examples/testmonitor/steps.py
```diff
@@ -0,0 +1,125 @@
...
+                text="My output string",
+                parameters=[
+                    Measurement(
+                        name="Temperature",
+                        status=Status.PASSED(),
+                        measurement="35",
+                        lowLimit="30",
+                        highLimit="40",
+                        units="C",
+                        comparisonType="Numeric",
```
rbell517, line 102: Whatever we end up doing with the extra properties we should make sure its shown in the examples. If we need a separate example for that then we should do that. In general I prefer more bite-sized examples that just show how one function works a certain way rather than a chain of examples that build on each other like you have, but both can work.

tests/integration/testmonitor/test_testmonitor.py
```diff
@@ -396,3 +435,461 @@ def __map_result_to_update_result_request(
...
+                    data=updated_data,
+                    inputs=updated_inputs,
+                    outputs=updated_outputs,
+                )
+            ]
+        )
+
+        assert update_response is not None
+        assert len(update_response.steps) == 1
+        assert update_response.steps[0].data == updated_data
```
rbell517, line 771: The data measurements are special and I would feel better if we broke out that assert into verifying each field to make sure the round trip serialization is working as expected. I'd rather not rely on the pydantic type comparison since we've got extras that I bet it ignores for equality.

tests/integration/testmonitor/test_testmonitor.py
```diff
@@ -396,3 +435,461 @@ def __map_result_to_update_result_request(
...
+                Measurement(
+                    name="Voltage",
+                    status="Passed",
+                    measurement="3.725",
+                    lowLimit="3.65",
+                    highLimit="3.8",
+                    units="V",
+                    comparisonType="GELE",
+                    specId="spec_01",
+                    specInfo={"specKey": "specValue"},
```
rbell517, line 748: These are only string-string dictionaries, so I'm not sure what this child object will do with your implementation. I guess covert that into a little json snippet?

tests/integration/testmonitor/test_testmonitor.py
```diff
@@ -396,3 +435,461 @@ def __map_result_to_update_result_request(
...
+        assert created_step.step_id == step_id
+        assert created_step.result_id == result_id
+        assert created_step.name == name
+        assert created_step.data == data
+        assert created_step.properties == properties
+        assert created_step.keywords == keywords
+        assert created_step.inputs == inputs
+        assert not created_step.outputs
+
+    def test__create_multiple_steps__multiple_creation_succeed(
```
rbell517, line 484: Would you do a version of this using the `children` field to make a multiple. You'd want to verify the parent is set properly on the children in the response since the response will be flat.

rbell517 (changes requested):

nisystemlink/clients/testmonitor/models/_step.py
SSadaiyappan-NI: yes @rbell517 , we are able to access the extra fields just like the defined ones

tests/integration/testmonitor/test_testmonitor.py
SSadaiyappan-NI: Based on the implementation, we are converting everything to string. 
```
specInfo="{'specKey': 'specValue'}"
```

tests/integration/testmonitor/test_testmonitor.py
SSadaiyappan-NI: added a new test

tests/integration/testmonitor/test_testmonitor.py
SSadaiyappan-NI: Created assert statements for individual fields and the extra fields as well

examples/testmonitor/steps.py
SSadaiyappan-NI: updated the examples

nisystemlink/clients/testmonitor/models/_step_data.py
SSadaiyappan-NI: both will work in our use case, since they both return known properties from the model. I have modified to use __fields__

examples/testmonitor/steps.py
```diff
@@ -20,7 +19,7 @@ def create_test_result():
             part_number="Example 123 AA",
             program_name="Example Name",
             host_name="Example Host",
-            status=Status.PASSED(),
+            status="Passed",
```
rbell517, line 22: Results should using the Status class. Measurements just have a string status, but results and steps have the more complex object.

test.py
rbell517, line 1: Revert this test file before you merge to master

rbell517 (approved):

test.py
SSadaiyappan-NI: Reverted the file. Did not notice it got committed. Thanks

examples/testmonitor/steps.py
SSadaiyappan-NI: Updated

PR #97, opened 2025-02-27
feat: Add and resolve query projection for products and specs respectively

shri2k2:

- Adds `projection` in the `query products` request and includes a separate response model, where all the fields are optional.
- Alters the request and response models for Products client, to resolve some inconsistencies.
- Includes a response model for `query specs` where all the fields of the model are made optional.

- Projection is something there in the actual backend service but missing in the product client. It has been included as a feature here.
- The projection for query specs works only if all the fields are optional in the response model. Previously, few of the fields were mandatory.
- The request models for Products Client were previously inconsistent. Those were fixed here, and this might be a breaking change.

- Integration tests are included.
- Manual testing has been done using a py script.


nisystemlink/clients/spec/models/_query_specs.py
```diff
@@ -1,9 +1,15 @@
...
+from nisystemlink.clients.spec.models._specification import (
+    Condition,
+    SpecificationCreation,
+    SpecificationLimit,
+    SpecificationType,
+    SpecificationUpdated,
+)
 
 
 class Projection(str, Enum):
```
SSSantosh18, line 15: Can we make this `SpecificationProjection` and orderby enum as `SpecificationOrderBy` to stay consistent with Product models?

If multiple clients (product and spec) are used in a script (which is possible), just having models with names like Projection and OrderBy will cause confusions

SSSantosh18 (approved):

nisystemlink/clients/product/models/_product_request.py
```diff
@@ -43,3 +31,18 @@ class Product(JsonModel):
 
     workspace: Optional[str]
     """The id of the workspace that this product belongs to."""
+
+
+class CreateProductRequest(BaseProductRequest):
```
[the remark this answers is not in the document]
ShriramS-Emerson: These are all the objects to be in place of the request body and not the request itself, as you are mentioning. But this is convention followed throughout the repo for other clients too.

examples/product/products.py
```diff
@@ -70,7 +71,7 @@ def create_some_products():
 
 # Query for just the ids of products that match the family
 values_query = QueryProductValuesRequest(
-    filter=f'family="{family}"', field=ProductField.ID
+    filter=f'family="{family}"', field=ProductProjection.ID
```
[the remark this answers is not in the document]
ShriramS-Emerson: Will change

nisystemlink/clients/spec/models/_query_specs.py
ShriramS-Emerson: Will update

nisystemlink/clients/product/models/_product_request.py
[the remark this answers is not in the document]
shri2k2: These are all the objects to be in place of the request body and not the request itself, as you are mentioning. But this is convention followed throughout the repo for other clients too.

examples/product/products.py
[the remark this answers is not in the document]
shri2k2: Yes, will change

nisystemlink/clients/product/models/_product_request.py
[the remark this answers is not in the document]
ShriramS-Emerson: @SSSantosh18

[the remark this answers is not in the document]
shri2k2: @SSSantosh18

nisystemlink/clients/product/_product_client.py
```diff
@@ -5,7 +5,6 @@
 from nisystemlink.clients import core
 from nisystemlink.clients.core._uplink._base_client import BaseClient
 from nisystemlink.clients.core._uplink._methods import delete, get, post
-from nisystemlink.clients.product.models import Product
 from uplink import Field, Query, returns
 
 from . import models
```
SSSantosh18, line 10: Please backoff retry for product client as included in the results client PR

nisystemlink/clients/spec/models/_query_specs.py
```diff
@@ -103,10 +109,74 @@ class QuerySpecificationsRequest(JsonModel):
...
+    keywords: Optional[List[str]] = None
+    """Keywords or phrases associated with the specification."""
+
+    properties: Optional[Dict[str, str]] = None
+    """Additional properties associated with the specification."""
+
+    """When the spec was created and when."""
+
+
 class QuerySpecifications(WithPaging):
```
SSSantosh18, line 176:
```suggestion
class PagedSpecifications(WithPaging):
```

SSSantosh18: to be consistent with other clients

SSSantosh18:
@rbell517 as changes in both Product and Spec client were related to projection and model fixes, they are breaking changes and we have combined them into a single PR. Let us know if there are any concerns with it.

Also, is there any specific process to do a major version bump? The angular commit documentation guideline related to schematic versioning in the contributing doc didn't mention anything about this.

SSSantosh18:
@rbell517, as we have created different models for Query response and Update request, @santhoshramaraj was suggesting if we could expose a method to map and create a update request model from query response so that it will be easy for the users

Maybe something like this within the existing model class?
```py
class UpdateProductRequest():
    def from(ProductResponse product): UpdateProductRequest
        update_request = ....logic to map model,,,
        return update_request
```

Please share your thoughts

nisystemlink/clients/product/_product_client.py
shri2k2: Added retry mechanism for both Specs and Product clients. Updated with the status codes as [408, 429, 502, 503, 504].

nisystemlink/clients/product/models/_product_response.py
```diff
@@ -0,0 +1,51 @@
+from datetime import datetime
+from typing import Dict, List, Optional
+
+from nisystemlink.clients.core._uplink._json_model import JsonModel
+
+
+class ProductResponse(JsonModel):
```
rbell517, line 7: I don't like this rename. The core response record should just be `Product`. All the fields should be Optional. It should be used for both the create response and the query response.

nisystemlink/clients/spec/models/_query_specs.py
```diff
@@ -103,10 +109,74 @@ class QuerySpecificationsRequest(JsonModel):
     """
 
 
-class QuerySpecifications(WithPaging):
+class SpecificationWithOptionalFields(SpecificationCreation, SpecificationUpdated):
```
rbell517, line 112: I don't like this as the query response type. The query should return a `Specification` with all fields optional. Instead of making a new type for the query response with all fields optional, we should make new update and create types that make fields required, not the other way around. The majority of the existing Specification type is already optional and you already have the SpecificationDefinition type that can serve this role for the create case, so I think this really just impacts the update request and response classes.

rbell517 (changes requested):

rbell517:
> @rbell517 as changes in both Product and Spec client were related to projection and model fixes, they are breaking changes and we have combined them into a single PR. Let us know if there are any concerns with it.
> 
> Also, is there any specific process to do a major version bump? The angular commit documentation guideline related to schematic versioning in the contributing doc didn't mention anything about this.

@SSSantosh18  @shri2k2 
To do a major version bump you add a trailing paragraph to your commit message starting with `BREAKING CHANGE:` that then describes the breaking change. It will cause the version bump to be major and add the breaking changes to the changelog. I will take an action to update the contributing docs to point to conventional commits instead of the angular page.

rbell517:
> @rbell517, as we have created different models for Query response and Update request, @santhoshramaraj was suggesting if we could expose a method to map and create a update request model from query response so that it will be easy for the users
> 
> Maybe something like this within the existing model class?
> 
> ```python
> class UpdateProductRequest():
>     def from(ProductResponse product): UpdateProductRequest
>         update_request = ....logic to map model....
>         return update_request
> ```
> 
> Please share your thoughts

@SSSantosh18 
I like this idea but lets save it for another PR. As stated in the results PR we will need to be careful when doing this w.r.t. workspace updates because they affect what permissions the update is required to have merely based on the presence of the workspace field in the request regardless of whether the workspace is actually changed. If you copy all of the fields you'll end up changing the required permissions unnecessarily. My suggestion in that PR was to add a `allow_workspace_update` field (default false) when doing a mapping like this to control whether to map over the workspace or not.

nisystemlink/clients/spec/models/_create_specs_request.py
```diff
@@ -22,14 +37,33 @@ class CreatedSpecification(
...
+
+    product_id: str
+    """Id of the product to which the specification will be associated."""
+
+    spec_id: str
+    """User provided value using which the specification will be identified.
+
+    This should be unique for a product and workspace combination.
+    """
+
```
SSadaiyappan-NI, line 59: these values are alerady inherited right? do we need them again

nisystemlink/clients/spec/models/_create_specs_request.py
```diff
@@ -22,14 +37,33 @@ class CreatedSpecification(
...
 
 class CreateSpecificationsPartialSuccess(JsonModel):
     """When some specs can not be created, this contains the list that was and was not created."""
 
     created_specs: Optional[List[CreatedSpecification]] = None
     """Information about the created specification(s)"""
 
-    failed_specs: Optional[List[SpecificationDefinition]] = None
+    failed_specs: Optional[List[CreateSpecificationsRequestObject]] = None
```
SSadaiyappan-NI, line 66: this should be `CreateSpecificationsRequest` right

nisystemlink/clients/spec/models/_create_specs_request.py
sam-rishi: Those fields are optional in the base model. For this create request, we need these fields as mandatory fields.

nisystemlink/clients/spec/models/_create_specs_request.py
sam-rishi: `failed_specs` is a list of failed `CreateSpecificationRequestObject` not `CreateSpecificationRequest`.  `CreateSpecificationRequest` has a list of `CreateSpecificationRequestObject`. The ones that fail among them ends up in this `failed_specs` list.

nisystemlink/clients/spec/models/_update_specs_request.py
```diff
@@ -5,14 +5,39 @@
...
+
+    spec_id: str
+    """User provided value using which the specification will be identified.
+
+    This should be unique for a product and workspace combination.
+    """
+
+    type: SpecificationType
+    """Type of the specification."""
+
```
SSadaiyappan-NI, line 36: lets make workspace mandatory

nisystemlink/clients/spec/models/_update_specs_request.py
```diff
@@ -23,13 +48,32 @@ class UpdatedSpecification(
 ):
     """A specification that was updated on the server."""
 
+    id: str
```
SSadaiyappan-NI, line 51: remove the inherited classes and add mandatory fields here

ShriramS-Emerson:
> > @rbell517 as changes in both Product and Spec client were related to projection and model fixes, they are breaking changes and we have combined them into a single PR. Let us know if there are any concerns with it.
> > Also, is there any specific process to do a major version bump? The angular commit documentation guideline related to schematic versioning in the contributing doc didn't mention anything about this.
> 
> @SSSantosh18 @shri2k2 To do a major version bump you add a trailing paragraph to your commit message starting with `BREAKING CHANGE:` that then describes the breaking change. It will cause the version bump to be major and add the breaking changes to the changelog. I will take an action to update the contributing docs to point to conventional commits instead of the angular page.

@rbell517 What is the suggested way of doing this?

Can we add a local commit with the `BREAKING CHANGE` message in it and then push it? Or should we update the commit message while trying to merge this branch? Or is there any better way of doing this?

shri2k2:
> > @rbell517 as changes in both Product and Spec client were related to projection and model fixes, they are breaking changes and we have combined them into a single PR. Let us know if there are any concerns with it.
> > Also, is there any specific process to do a major version bump? The angular commit documentation guideline related to schematic versioning in the contributing doc didn't mention anything about this.
> 
> @SSSantosh18 @shri2k2 To do a major version bump you add a trailing paragraph to your commit message starting with `BREAKING CHANGE:` that then describes the breaking change. It will cause the version bump to be major and add the breaking changes to the changelog. I will take an action to update the contributing docs to point to conventional commits instead of the angular page.

@rbell517 What is the suggested way of doing this?

Can we update the commit message while trying to merge this branch? Or is there any better way of doing this?

nisystemlink/clients/spec/models/_specification.py
```diff
@@ -77,7 +77,7 @@ class SpecificationDefinition(SpecificationUserManaged):
     category: Optional[str] = None
     """Category of the specification."""
 
-    type: SpecificationType
+    type: Optional[SpecificationType] = None
```
SSSantosh18, line 80: this is required field in both create and update responses, so this can be moved to Specification

nisystemlink/clients/spec/models/_create_specs_request.py
```diff
@@ -5,31 +5,67 @@
...
 
-    specs: Optional[List[SpecificationDefinition]] = None
+    specs: Optional[List[CreateSpecificationsRequestObject]] = None
     """List of specifications to be created."""
 
 
-class CreatedSpecification(
-    SpecificationServerManaged, SpecificationUserManaged, SpecificationCreation
-):
+class CreatedSpecification(SpecificationCreation):
```
SSSantosh18, line 33: SpecificationCreation has created** fields as optional. But they are required

SSSantosh18, line 33: Both create and update response have same parameters except for created** and updated** fields. We can create `BaseSpecificationRequest` with the common parameters and derive create and update response models from this model

nisystemlink/clients/spec/models/_update_specs_request.py
```diff
@@ -4,32 +4,76 @@
...
 from nisystemlink.clients.spec.models._specification import (
     Specification,
-    SpecificationServerManaged,
+    SpecificationType,
     SpecificationUpdated,
-    SpecificationUserManaged,
 )
 
 
+class UpdateSpecificationsRequestObject(Specification):
```
SSSantosh18, line 12: anyway we are defining the required fields again. So, SpecificationDefinition can be used as base class instead of Specification

nisystemlink/clients/product/models/_product.py
```diff
@@ -5,12 +5,18 @@
 
 
 class Product(JsonModel):
-    """Contains information about a product."""
+    """Contains information about a product, where all the fields are optional.
+
+    - This is because when using query products' projection, user can request for any of the available
```
SSSantosh18, line 10:
```suggestion
    - This is because when using query products' projection, specific product fields alone can be queried.
```

nisystemlink/clients/product/models/_product.py
```diff
@@ -5,12 +5,18 @@
...
+    fields. So, we are making sure that all the available fields are optional.
```
SSSantosh18, line 11:
```suggestion
```

nisystemlink/clients/product/models/_product.py
```diff
@@ -5,12 +5,18 @@
...
+
+    - Also while creating a product it is not mandatory that the product should be created with all the fields.
```
SSSantosh18, line 13:
```suggestion
```

nisystemlink/clients/spec/models/_create_specs_request.py
```diff
@@ -2,34 +2,54 @@
...
     """List of specifications to be created."""
 
 
-class CreatedSpecification(
-    SpecificationServerManaged, SpecificationUserManaged, SpecificationCreation
-):
+class CreatedSpecification(BaseSpecificationResponse, SpecificationCreation):
     """A specification successfully created on the server."""
 
+    workspace: Optional[str] = None
```
SSSantosh18, line 39: this is not optional

SSSantosh18: it can be directly added to the BaseSpecificationResponse as required

nisystemlink/clients/spec/models/_create_specs_request.py
SSSantosh18: Swagger might be incorrect here. Created** fields for create response and Updated** fields for update response will come in response always

nisystemlink/clients/spec/models/_update_specs_request.py
```diff
@@ -2,34 +2,68 @@
...
     """List of specifications to be updated."""
 
 
-class UpdatedSpecification(
-    SpecificationUserManaged,
-    SpecificationServerManaged,
-    SpecificationUpdated,
-):
+class UpdatedSpecification(BaseSpecificationResponse, SpecificationUpdated):
     """A specification that was updated on the server."""
```
SSSantosh18, line 53: make updated** fields required here as suggested in other comment

nisystemlink/clients/spec/models/_base_specification_response.py
```diff
@@ -0,0 +1,24 @@
+from nisystemlink.clients.core._uplink._json_model import JsonModel
+
+
+class BaseSpecificationResponse(JsonModel):
```
SSSantosh18, line 4: this could be within _create_specs_request.py itself

SSSantosh18 (approved):

nisystemlink/clients/spec/models/_query_specs.py
```diff
@@ -103,10 +107,16 @@ class QuerySpecificationsRequest(JsonModel):
...
+    Specification, SpecificationCreation, SpecificationUpdated
+):
+    """A full specification with update and create history."""
+
+
+class PagedSpecifications(WithPaging):
     """The list of matching specifications and a continuation token to get the next items."""
 
-    specs: Optional[List[SpecificationWithHistory]] = None
+    specs: Optional[List[QuerySpecificationsResponse]] = None
```
SSSantosh18, line 119: We can make this Specification as suggested by Richard if we can make Specification also derive from SpecificationCreation and SpecificationUpload.

SSSantosh18: With the current usage of models, Specification will only be used by QueryResponse. So, this should be ok

nisystemlink/clients/spec/models/_query_specs.py
SSSantosh18: @rbell517 I have reset your vote now though there are still some minor comments to be fixed to avoid further delays in confirming the model structure. Existing comments once resolved should modify the models as suggested by you. 

1. For PagedSpecifications, `Specification` will be used
2. `BaseSpecificationResponse` containing common fields of `CreatedSpecification` and `UpdatedSpecification` which will be used in partial success response
3. Create and update request models deriving from SpecificationDefinition.

This will be new structure. Please let us know if you have different thoughts

nisystemlink/clients/product/models/_product.py
rbell517: I would probably just drop the notice that all the fields are optional. If we want an explanation it would be better to give the context on when values could be `None`. For example, it can be None if no projection is applied and the field is unset, or it can be None when the field is not included in the projection.

nisystemlink/clients/product/_product_client.py
```diff
@@ -30,7 +32,7 @@ def __init__(self, configuration: Optional[core.HttpConfiguration] = None):
 
     @post("products", args=[Field("products")])
     def create_products(
-        self, products: List[Product]
+        self, products: List[models.CreateProductRequest]
```
rbell517, line 35: I'd rather update the import instead of including `models` here.

nisystemlink/clients/spec/models/_query_specs.py
rbell517: Yes Specification should include all fields returned by a query, which includes the created* and updated* fields

nisystemlink/clients/spec/models/_specification.py
```diff
@@ -41,10 +41,10 @@ class SpecificationType(Enum):
 
 
 class SpecificationUserManaged(JsonModel):
```
rbell517, line 43: Do we still need this separate from SpecificationDefinition now that you have a separate create object? I don't see a use of it, and I'd rather remove it.

nisystemlink/clients/spec/models/_specification.py
```diff
@@ -108,6 +105,9 @@ class SpecificationDefinition(SpecificationUserManaged):
 class Specification(SpecificationDefinition, SpecificationServerManaged):
     """The complete definition of a specification."""
 
+    type: Optional[SpecificationType] = None
```
rbell517, line 108: Is there a reason we can't add this `type` field to SpecificationDefinition? We override it to be required in the update and create requests already.

nisystemlink/clients/spec/models/_update_specs_request.py
```diff
@@ -2,34 +2,68 @@
...
     SpecificationUpdated,
-    SpecificationUserManaged,
 )
 
 
+class UpdateSpecificationsRequestObject(SpecificationDefinition):
+    id: str
+    """The global Id of the specification."""
+
+    version: int
```
rbell517, line 19: Order the fields alphabetically so they match the swagger

rbell517 (changes requested):

nisystemlink/clients/spec/models/_query_specs.py
rbell517: Yes what you have is good with me. I'm not so zealous about making base classes for the common fields like you're doing with BaseSpecificationResponse, but its reasonable here.

nisystemlink/clients/spec/models/_specification.py
sam-rishi: Included the fields inside SpecificationDefinition and removed SpecificationUserManaged.

nisystemlink/clients/spec/models/_specification.py
sam-rishi: Moved `type` to SpecificationDefinition.

nisystemlink/clients/spec/models/_update_specs_request.py
sam-rishi: Ordered the fields to match the swagger order.

nisystemlink/clients/product/_product_client.py
shri2k2: This is how it is followed in all the other clients. Like importing `from . import models` and then accessing all the required models from it.

nisystemlink/clients/product/models/_product.py
SSSantosh18: @shri2k2, I would also prefer the notice about optional fields. If we just say it would be none when the field is unset, it doesn't apply to all fields because some fields are required fields of product. So, this would lead to confusions.

Lets avoid that

rbell517 (approved):

PR #98, opened 2025-03-04
feat: Export Specs as Dataframe

sam-rishi:

- Adds utility for exporting specs as a dataframe.
- User can pass `column projection` and a function for `condition formatting` in the parameters.
- If `column projection` is not passed, all the columns will be included in the dataframe as a default approach.
- If `condition formatting` is not passed, default condition formatting will be done. In the default condition formatting, the column headers will be in `condition_conditionName(conditionUnit)` format if the condition value is `NumericConditionValue` and `condition_conditionName` format if the condition value is `StringConditionValue`. The column value will be in `[min: num; max; step], num, num` format where data within `[]` is condition range and other num values are discrete values if the condition value is `NumericConditionValue` and `str, str, str` where the str values are discrete values if the condition value is `StringConditionValue`.
- If `column projection` is passed and either `CONDITION_NAME` or `CONDITION_VALUES` is not included, the `condition formatting` won't be done and conditions columns will be excluded in the dataframe.

- This utility can be used to export specs as a dataframe with only the specified columns and specified condition format.
- The user can easily retrieve only the necessary data in their desired format.

- Automated integration tests are included.


sam-rishi:
@rbell517 @santhoshramaraj 
- NOTE: Only the following files need to be reviewed in this PR:
  - nisystemlink/clients/spec/utilities/`__init__.py`
  - nisystemlink/clients/spec/uitilities/`_client_utilities.py`
  - nisystemlink/clients/spec/uitilities/`_constants.py`
  - nisystemlink/clients/spec/uitilities/`_dataframe_utilities.py
  - `poetry.lock`
  - `poetry.toml`
  - test/integration/spec/`test_spec.py`
- Other changes are already present in the `feat: Add and resolve query projection for products and specs respectively` PR (https://github.com/ni/nisystemlink-clients-python/pull/97).

nisystemlink/clients/spec/utilities/_dataframe_utilities.py
```diff
@@ -0,0 +1,319 @@
...
+                          This is an optional parameter. By default column header will be
+                          "condition_<conditionName>(<conditionUnit>)".
+                          The column value will be "[min: num; max: num, step: num], num, num"
+                          where data within the '[]' is numeric condition range and other num
+                          values are numeric condition discrete values.
+                          The column value will be "str, str, str" - where str values are the
+                          condition discrete values for a string condition.
+
+    Returns:
+        The list of specs of the specified product as a dataframe.
```
rbell517, line 72: The format of the dataframe should be detailed here, not just in the condition_format doc comment

nisystemlink/clients/spec/utilities/_dataframe_utilities.py
```diff
@@ -0,0 +1,319 @@
...
+    column_projection: Optional[List[SpecificationProjection]] = None,
+    condition_format: Callable[
+        [List[Condition]], Dict[str, str]
+    ] = __serialize_conditions,
+) -> pd.DataFrame:
+    """Get specs of a specific product as a dataframe.
+
+    Args:
+        client: The Spec Client to use for the request.
+        product_ids: ID of the product to query specs.
```
rbell517, line 52:
```suggestion
        product_id: ID of the product to query specs.
```

nisystemlink/clients/spec/utilities/_dataframe_utilities.py
```diff
@@ -0,0 +1,319 @@
...
+        __generate_condition_column_header(condition): ", ".join(
+            __get_condition_values(condition)
+        )
+        for condition in conditions
+    }
+
+
+def get_specs_dataframe(
+    client: SpecClient,
+    product_id: str,
```
rbell517, line 41: The underlying query API supports multiple products. Is there a reason you'd only want to support one in this function?

nisystemlink/clients/spec/utilities/_dataframe_utilities.py
```diff
@@ -0,0 +1,319 @@
...
+    return {
+        __generate_condition_column_header(condition): ", ".join(
+            __get_condition_values(condition)
+        )
+        for condition in conditions
+    }
+
+
+def get_specs_dataframe(
+    client: SpecClient,
```
rbell517, line 40: Can you give some justification for why you want this utility to include the query part? I agree we need to invest in some helpers for paging to get all the records, but otherwise that work is pretty separate from the work to convert the return specifications into a data frame. You're not doing anything fancy like building the data frame from each page of results while waiting for the next page. My inclination is that we would want a function that takes a list of Specifications and converts that into a dataframe, which is basically what your `__serialize_specs` function does.

If we really do want to include the query in the single function, I'd also like us to consider adding the function to produce a dataframe directly to the client e.g. query_specs_as_dataframe. Dataframes can be concated or merged fairly easily, so its not like we need to retrieve all the pages before we touch pandas. Putting this on the client could have some encapsulation value if we think there should be a standard way to query for specs in such a way that would product a dataframe efficiently. That might not be very interesting for specs, but for TestMonitor where we need to query results and steps, and we have options to use CSV instead of JSON responses, that becomes more interesting. We have room in the services to invest in adding different response types like Arrow that we could then swap to without any breaking changes (though if we are using Arrow we'd prefer using PyArrow instead of pandas).

I want an answer that is going to work for specs and test monitor, so lets make sure to discuss our options. If we need a meeting to hash it out please get it on the calendar. You can check with Usha if the Thursday tech sync is open, though I only have the first half hour available.

tests/integration/__init__.py
SSadaiyappan-NI, line 1: why are we deleting this file?

nisystemlink/clients/product/models/_product.py
SSadaiyappan-NI, line 1: This change is not relevant to this PR, please remove it

nisystemlink/clients/spec/utilities/_constants.py
```diff
@@ -0,0 +1,5 @@
+CONDITION_COLUMN_HEADER_PREFIX = "condition_"
```
SSadaiyappan-NI, line 1: we can have a class for the headers, probably something like DataFrameHeaders

nisystemlink/clients/spec/utilities/_dataframe_utilities.py
```diff
@@ -0,0 +1,248 @@
...
+    Returns:
+        Conditions as a dictionary.
+    """
+    return {
+        __generate_condition_column_header(condition): ", ".join(
+            __get_condition_values(condition)
+        )
+        for condition in conditions
+        if condition.name and condition.value
+    }
```
SSadaiyappan-NI, line 32: the private function should go below the public function

nisystemlink/clients/spec/utilities/_dataframe_utilities.py
```diff
@@ -0,0 +1,248 @@
...
+        header for header in column_headers if __is_standard_column_header(header)
+    ]
+    condition_headers = [header for header in column_headers if "condition_" in header]
+    properties_headers = [
+        header for header in column_headers if "properties." in header
+    ]
+    formatted_column_headers += (
+        condition_headers
+        + properties_headers
+        + (["keywords"] if "keywords" in column_headers else [])
```
SSadaiyappan-NI, line 112: keywords can come before properties

SSadaiyappan-NI, line 112: use the methods created to check for headers

tests/unit/spec/test_spec_dataframe_utilitites.py
```diff
@@ -0,0 +1,244 @@
...
+            symbol="Range with required value",
+            block="Amplifier",
+            limit=SpecificationLimit(min=1.2, max=1.5),
+            unit="mV",
+        ),
+    ]
+
+    return specs
+
+
```
SSadaiyappan-NI, line 107: mark it as a unit test

tests/unit/spec/test_spec_dataframe_utilitites.py
```diff
@@ -0,0 +1,244 @@
...
+                {
+                    key: value
+                    for key, value in vars(spec).items()
+                    if key not in ["conditions"]
+                }
+            )
+        expected_specs_df = pd.json_normalize(specs_dict)
+        expected_specs_df.dropna(axis="columns", how="all", inplace=True)
+        keywords_column = expected_specs_df.pop("keywords")
+        expected_specs_df["keywords"] = keywords_column
```
SSadaiyappan-NI, line 123: why are we doing this?

tests/unit/spec/test_spec_dataframe_utilitites.py
```diff
@@ -0,0 +1,244 @@
...
+
+        specs_df = convert_specs_to_dataframe(specs=specs)
+        specs_df = specs_df.drop(
+            columns=[
+                "condition_Temperature(C)",
+                "condition_Supply Voltage(mV)",
+                "condition_Package",
+            ]
+        )
+
```
SSadaiyappan-NI, line 133: are we not going to be testing these columns?

tests/unit/spec/test_spec_dataframe_utilitites.py
```diff
@@ -0,0 +1,244 @@
...
+            "name",
+            "category",
+            "type",
+        ]
+
+        specs_df = convert_specs_to_dataframe(specs=specs)
+        specs_columns = specs_df.columns.to_list()
+
+        assert not specs_df.empty
+        assert specs_columns == expected_specs_columns
```
SSadaiyappan-NI, line 244: assert aleast data of some columns

tests/unit/spec/test_spec_dataframe_utilitites.py
```diff
@@ -0,0 +1,244 @@
...
+            type=SpecificationType.PARAMETRIC,
+            symbol="Range with required value",
+            block="Amplifier",
+            limit=SpecificationLimit(min=1.2, max=1.5),
+            unit="mV",
+        ),
+    ]
+
+    return specs
+
```
SSadaiyappan-NI, line 106: can we create an expected dataframe structure like this and use it in other places
image

tests/unit/spec/test_spec_dataframe_utilitites.py
```diff
@@ -0,0 +1,244 @@
...
+            columns=[
+                "condition_Temperature(C)",
+                "condition_Supply Voltage(mV)",
+                "condition_Package",
+            ]
+        )
+
+        assert not specs_df.empty
+        assert len(specs_df) == 3
+        assert specs_df.equals(expected_specs_df), specs_df.compare(expected_specs_df)
```
SSadaiyappan-NI, line 136:
```
pd.testing.assert_frame_equal(products_dataframe, expected_products_dataframe, check_dtype=True)
```
use inbuilt pandas method to assert dataframes

SSadaiyappan-NI (changes requested):

tests/integration/__init__.py
sam-rishi: Deleted by mistake, fixed it.

nisystemlink/clients/product/models/_product.py
sam-rishi: It was added when I pulled `fix-query-projection` branch before it was merged to master. Fixed it.

nisystemlink/clients/spec/utilities/_dataframe_utilities.py
sam-rishi: This function is used as a default value in the public method's parameter. So it must be defined before that method, or else it won't compile.

tests/unit/spec/test_spec_dataframe_utilitites.py
sam-rishi: It's to move keywords column to the last. Now I changed it to move keywords column to the be before properties.

tests/unit/spec/test_spec_dataframe_utilitites.py
sam-rishi: Columns are tested in the below method.

tests/unit/spec/test_spec_dataframe_utilitites.py
```diff
@@ -0,0 +1,269 @@
...
+            limit=SpecificationLimit(min=1.2, max=1.5),
+            unit="mV",
+        ),
+    ]
+
+    return specs
+
+
+@pytest.fixture
+def expected_specs_dataframe(specs) -> pd.DataFrame:
```
SSadaiyappan-NI, line 109: can we reuse this method somehow for all test cases, like provide the expected dataframe based on the inputs specs. and it will normalize it and drop all na columns. Finally we should be able to assert this to actual dataframe. Currently its only used in one test

sam-rishi: Modified it into a reusable method.

nisystemlink/clients/spec/utilities/_dataframe_utilities.py
```diff
@@ -0,0 +1,246 @@
...
+    return {
+        __generate_condition_column_header(condition): ", ".join(
+            __get_condition_values(condition)
+        )
+        for condition in conditions
+        if condition.name and condition.value
+    }
+
+
+def convert_specs_to_dataframe(
```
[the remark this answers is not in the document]
SSSantosh18: all three combined only make a unique combination

SSSantosh18, line 31: we need to create an example for this similar to how we create example for client APIs as we are exposing a callback function. We can have the use the same code of our default condition format as the example too

nisystemlink/clients/spec/utilities/_dataframe_utilities.py
```diff
@@ -0,0 +1,246 @@
...
+                          "condition_<conditionName>(<conditionUnit>)".
+                          The column value will be "[min: num; max: num, step: num], num, num"
+                          where data within the '[]' is numeric condition range and other num
+                          values are numeric condition discrete values.
+                          The column value will be "str, str, str" - where str values are the
+                          condition discrete values for a string condition. If the condition doesn't
+                          have values, it will not be added to the dataframe.
+
+    Returns:
+        The list of specs of the specified product as a dataframe. Condition column will be formatted based on the
```
SSSantosh18, line 59:
```suggestion
        A Pandas DataFrame with the each spec fields having a separate column. Following fields are split into sub-columns.
        - limit: Split into limit.min, limit.typical and limit.max columns.
        - conditions: format of the condition columns are decided by the  `condition_format` argument of this function
        - Properties: All the unique properties across all specs will be split into separate columns. For example,
        properties.property1, properties.property2, etc.
```

tests/spec/test_spec_dataframe_utilitites.py
```diff
@@ -0,0 +1,248 @@
...
+                Condition(
+                    name="Package",
+                    value=StringConditionValue(
+                        condition_type=ConditionType.STRING,
+                        discrete=["D", "QFIN"],
+                    ),
+                ),
+            ],
+            keywords=["Test specification only", "First"],
+            properties={"Comments": "comma separated with unicode"},
```
SSSantosh18, line 54: we can have multiple properties in the next spec, where one is Comments and other is a different property

tests/integration/spec/test_spec.py
```diff
@@ -17,6 +17,7 @@
     SpecificationLimit,
     SpecificationProjection,
     SpecificationType,
+    StringConditionValue,
```
SSSantosh18, line 20: why is this file changed?

nisystemlink/clients/spec/utilities/_dataframe_utilities.py
```diff
@@ -0,0 +1,246 @@
...
+from nisystemlink.clients.spec.models._condition import (
+    Condition,
+    NumericConditionValue,
+    StringConditionValue,
+)
+from nisystemlink.clients.spec.models._specification import Specification
+from nisystemlink.clients.spec.utilities._constants import DataFrameHeaders
+
+
+def __serialize_conditions(conditions: List[Condition]) -> Dict[str, str]:
```
SSSantosh18, line 13: we can move this method below and public method and see if we can use forward referencing of this method if method reference is what causing error in moving this down in the file

tests/integration/spec/test_spec.py
sam-rishi: It was because tests for dataframe utilities were in this file before. Removed the changes.

nisystemlink/clients/spec/utilities/_dataframe_utilities.py
sam-rishi: Moved the private method below and removed it from default arguments. Using it inside the method directly.

nisystemlink/clients/spec/utilities/_dataframe_utilities.py
SSSantosh18: As discussed over a call, removed the batching utility for now, which will be later implemented as a generic function for all clients. Also, we will try out using generator and reduce in our notebook to improve batching mem utilization and performance. This prototype will be shared later and based on which this generic function can be designed and created.

Now, the dataframe utility will just take list[specs] and convert it into dataframe. The same would be done for the TestMonitor utilities too

nisystemlink/clients/spec/utilities/_dataframe_utilities.py
```diff
@@ -0,0 +1,243 @@
...
+    }
+
+
+def convert_specs_to_dataframe(
+    specs: List[Specification],
+    condition_format: Optional[
+        Callable[[List[Condition]], Dict]
+    ] = __serialize_conditions,
+) -> pd.DataFrame:
+    """Format specs with respect to the provided condition format.
```
SSSantosh18, line 37:
```suggestion
    """Creates a Pandas DataFrame for the results
```

nisystemlink/clients/spec/utilities/_dataframe_utilities.py
```diff
@@ -0,0 +1,243 @@
...
+
+    Args:
+        specs: List of specs of the specified products.
```
SSSantosh18, line 40:
```suggestion
        specs: List of specs.
```

nisystemlink/clients/spec/utilities/_dataframe_utilities.py
```diff
@@ -0,0 +1,243 @@
...
+
+
+def __serialize_numeric_condition_range(value: NumericConditionValue) -> List[str]:
+    """Serialize ranges of a numeric condition value.
+
+    Args:
+        value: A condition's value with NumericConditionValue type.
+
+    Returns:
+        The list of ranges of the given value in a specific format.
```
SSSantosh18, line 217:
```suggestion
        The list of ranges of the given condition where each range will be in string format `[min: <value>; max: <value>; step: <value>]` if the corresponding fields are not none
```

tests/spec/test_spec_dataframe_utilitites.py
```diff
@@ -0,0 +1,278 @@
...
+        ),
+    ]
+
+    return specs
+
+
+@pytest.mark.enterprise
+@pytest.mark.unit
+class TestSpecDataframeUtilities:
+    def test__convert_specs_to_dataframe__returns_specs_dataframe(self, specs):
```
SSSantosh18, line 112: combine first two tests

tests/spec/test_spec_dataframe_utilitites.py
```diff
@@ -0,0 +1,290 @@
...
+                    },
+                    **(conditions_dict[index] if conditions_dict else {}),
+                    **(keywords_dict[index] if conditions_dict else {}),
+                    **(properties_dict[index] if conditions_dict else {}),
+                }
+            )
+            index += 1
+        expected_specs_df = pd.DataFrame(specs_dict)
+        expected_specs_df.dropna(axis="columns", how="all", inplace=True)
+        print(expected_specs_df)
```
SSSantosh18, line 287: please remove your debuggers😂

SSSantosh18 (approved):

nisystemlink/clients/spec/utilities/_dataframe_utilities.py
rbell517: Examples for the other data frames would be useful, too, but it doesn't need to be in the same PR as the functionality if that is easier for your team.

nisystemlink/clients/spec/utilities/_dataframe_utilities.py
```diff
@@ -0,0 +1,245 @@
...
+    """
+    if not condition.value:
+        return []
+
+    values = []
+
+    if isinstance(condition.value, NumericConditionValue):
+        values.extend(__serialize_numeric_condition_range(value=condition.value))
+
+    values.extend(__serialize_condition_discrete_values(value=condition.value))
```
rbell517, line 205: Why isn't keeping Condition as an object preferred to the string representation? What do clients intend to do with conditions where a string representation is useful? Aren't they just going to need to de-serialize it back into something useful? Or we want a degree of flattening, why not leave the range and discrete fields as lists and 

I'm also wondering if the better approach would be to denormalize based on condition, so for each condition in the spec you'd get a separate dataframe row. Or even further would be a separate row for each condition value, including elaborating the range steps. 

What is most useful will depend on the calculation we ultimately need this data frame for. I think the most common would be for test coverage and limit deviation, right? These would ultimately mean producing a separate row per condition _combination_ on the spec, expanding any range conditions into their discrete values. 

Next most common would be looking at limits or other spec level data independent on conditions, and what you have now is a reasonable start to supporting that, but I still question serializing into a string as the default behavior. 

Here are the changes I recommend:
- Change the formatting callable to give all the conditions and allow returning a list of dicts, where each entry is a row (key is column, value is value). This is equivalent to a `flatMap` in the map-reduce parlance. A generator could also be a good approach here. We should also consider returning a sub-dataframe for the spec-conditions that we would then join with the parent spec dataframe.
- Create standard mappers to produce a standard set of data frames:
   - A data frame with spec-condition per row that has a name column but leaves the condition value as `NumericConditionValue | StringConditionValue` object
   - A data frame with spec per row that leaves conditions as an array of `Condition`. The mapping here would be the simplest, and would be a reasonable default. If the caller wants they can then use `pd.explode` to convert that conditions array however they want. 
   - A data frame with spec-condition-combination per row that expands ranges to discrete values. This is the case where using pandas to make a sub-dataframe is probably easiest.

Also note we have a similar (sans-combinations) problem for steps and measurements, so whatever we do here we'd want a similar approach there.

rbell517 (changes requested):

tests/spec/test_spec_dataframe_utilitites.py
```diff
@@ -0,0 +1,288 @@
...
+
+        assert not specs_df.empty
+        assert len(specs_df) == 3
+        pd.testing.assert_frame_equal(specs_df, expected_specs_df, check_dtype=True)
+        assert specs_df["created_at"].dtype == "datetime64[ns, UTC]"
+        assert specs_df["updated_at"].dtype == "datetime64[ns, UTC]"
+        assert specs_df["type"].dtype == "object"
+        assert isinstance(specs_df["type"].iloc[0], SpecificationType)
+        assert specs_df["limit"].dtype == "object"
+        assert isinstance(specs_df["limit"].iloc[0], SpecificationLimit)
```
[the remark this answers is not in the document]
sam-rishi: Added columns for `limit.min`, `limit.max` and `limit.typical`.

tests/spec/test_spec_dataframe_utilitites.py
```diff
@@ -0,0 +1,288 @@
...
+
+        specs_df = convert_specs_to_dataframe(specs=specs)
+
+        assert not specs_df.empty
+        assert len(specs_df) == 3
+        pd.testing.assert_frame_equal(specs_df, expected_specs_df, check_dtype=True)
+        assert specs_df["created_at"].dtype == "datetime64[ns, UTC]"
+        assert specs_df["updated_at"].dtype == "datetime64[ns, UTC]"
+        assert specs_df["type"].dtype == "object"
+        assert isinstance(specs_df["type"].iloc[0], SpecificationType)
```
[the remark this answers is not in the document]
sam-rishi: Translated the enum into it's string equivalent, as we won't be have any further use case for it.

nisystemlink/clients/spec/utilities/_dataframe_utilities.py
```diff
@@ -0,0 +1,245 @@
...
+    """Creates a Pandas DataFrame for the specs.
+
+    Args:
+        specs: List of specs.
+        condition_format: Function which takes in a list of condition objects and returns
+                          a dictionary of condition and its values. The dictionary keys
+                          should be the condition name and the values should be the condition
+                          value in any format you need. Keys will be used as the dataframe
+                          column header and values will be used as the row cells for the
+                          respective column header. If not passed, default condition format will be used.
```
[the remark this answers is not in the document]
sam-rishi: Yes, if none is passed conditions won't be included. Added it to the docstring.

tests/spec/test_spec_dataframe_utilitites.py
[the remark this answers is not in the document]
SSSantosh18: It can be translated to String as that is the required info, not much processing is going to happen with the enum dtype after getting df using this utility

nisystemlink/clients/spec/utilities/_dataframe_utilities.py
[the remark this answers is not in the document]
rbell517: We didn't discuss it in the meeting this morning, but what you do with the conditions mapping function would impact what you'd want the index to be. Its also not clear to me with this parameter if that would mean you're creating a MultiIndex index or a new column that is the serialized combination of these columns. I would expect a MultiIndex would be more useful. Either way doing this would impact your data access in ways that probably would not help usability. Workspace and product IDs are not very commonly known, compared with the their name and part number equivalents.

I'm inclined to leave this unindexed or indexed by `id` if it is available in the projection (not `specId`) and let the caller reindex in whatever way makes sense for them.

rbell517 (changes requested):

nisystemlink/clients/spec/utilities/_dataframe_utilities.py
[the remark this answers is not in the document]
SSSantosh18: Unindexed is fine if the multi-index might not be that useful. Indexing by Id of spec might also not be that useful as it is not generally used that much in other scenarios. For eg, for mapping a spec to a step, SpecId is generally used by users instead of id.

nisystemlink/clients/spec/utilities/_dataframe_utilities.py
SSSantosh18: As, aligned in the call for the starter we would bring some level of reusability based on the suggestions as follows

### Specs

1. Condition formatter will take list of all conditions under a spec and give list[dict] where each entry is separate row
2. Default condition will be returned as a object in single row
3. We will also expose the method that formats condition into readable string
       a. Users can pass this as callback if they want the condition as string

### Meas data

1. All measurement will be in separate row
2. By default only parameters with both "name" and "value" keys are considered as measurement
      a. all non measurement parameters will be in filtered out and will not be present in the dataframe
      b. A callback function will be provided which will take in Dict and return bool, this can be used to add logic on deciding whether a parameter is measurement or not
      c. A method with default logic will be assigned as the default callback function
      d. If callback set as none, all params are split into separate rows

nisystemlink/clients/spec/utilities/_dataframe_utilities.py
```diff
@@ -0,0 +1,323 @@
...
+        conditions: List of all conditions in a spec.
+
+    Returns:
+        Conditions as a list of dictionary. The column header will be
+        "condition_<conditionName>(<conditionUnit>)".
+        The column value will be "[min: num; max: num, step: num], num, num"
+        where data within the '[]' is numeric condition range and other num
+        values are numeric condition discrete values.
+        The column value will be "str, str, str" - where str values are the
+        condition discrete values for a string condition. If the condition doesn't
```
SSSantosh18, line 33:
```suggestion
        condition discrete values for a string condition. If the condition doesn't
        have a name and value, it will be skipped
```

tests/spec/test_spec_dataframe_utilitites.py
```diff
@@ -250,39 +262,144 @@ def test__convert_specs_to_dataframe_without_condition_values__returns_specs_dat
...
+        assert specs_df["keywords"].dtype == "object"
+        assert isinstance(specs_df["keywords"].iloc[0], List)
+
+    def test__serialize_conditions_to_string__returns_only_conditions_with_value_in_string_format(
+        self, specs
+    ):
+        expected_conditions_dict = [
+            [
+                {
+                    "condition_Temperature(C)": "[min: -25.0; max: 85.0; step: 20.0], 1.3, 1.5, 1.7",
```
SSSantosh18, line 343: we should also have some conditions where there are multiple ranges and where some range have step or min/max as none

nisystemlink/clients/spec/utilities/_dataframe_utilities.py
```diff
@@ -63,16 +84,26 @@ def convert_specs_to_dataframe(
             - Properties: All the unique properties across all specs will be split into separate columns.
             For example, properties.property1, properties.property2, etc.
     """
+    if not condition_format:
+        condition_format = __default_condition_formatting
```
SSSantosh18, line 88:
```suggestion
        condition_format = __default_condition_formatting if not condition_format
```

SSSantosh18: will this work?

sam-rishi: Modified it to `condition_format = condition_format or __default_condition_formatting`

nisystemlink/clients/spec/utilities/_dataframe_utilities.py
```diff
@@ -0,0 +1,323 @@
...
+        header for header in column_headers if __is_property_header(header=header)
+    ]
+    formatted_column_headers = (
+        standard_column_headers
+        + condition_headers
+        + (["keywords"] if "keywords" in column_headers else [])
+        + properties_headers
+    )
+
+    return specs_dataframe.reindex(columns=formatted_column_headers)
```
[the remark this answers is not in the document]
sam-rishi: Changed into copy=False.

nisystemlink/clients/spec/utilities/_dataframe_utilities.py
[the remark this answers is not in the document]
rbell517: I'm no pandas expert but as far as I can tell you do not need to join by index, you can join (`merge` technically) by any column. Presumably it is faster if you join by index, and it looks like the index is maintained across the join (as a MultiIndex) which can have some value.

nisystemlink/clients/spec/utilities/_dataframe_utilities.py
```diff
@@ -100,6 +95,53 @@ def convert_specs_to_dataframe(
...
+            for key, value in vars(spec).items()
+            if key not in ["type", "limit", "conditions"]
+        },
+        **(__serialize_type(spec.type) if spec.type else {}),
+        **(__serialize_limits(spec.limit) if spec.limit else {}),
+        **{key: value for key, value in condition.items()},
+    }
+
+
+def __default_condition_formatting(
```
rbell517, line 122: Similar to the summarize case, our out of the box formatting should be public functions. This allows the caller to explicitly and declaratively opt in to the default or take the default function as a baseline they work on top of. That should also mean we give this a more descriptive name, like `condition_per_row`

nisystemlink/clients/spec/utilities/_dataframe_utilities.py
```diff
@@ -100,6 +95,53 @@ def convert_specs_to_dataframe(
...
+        conditions: List of all conditions in a spec.
+
+    Returns:
+        Conditions as a list of dictionary. The key will be
+        the condition name and the value will be the condition value which is
+        either Numeric Condition Value, String Condition Value or None.
+    """
+    return [
+        {
+            condition.name: condition.value
```
rbell517, line 138: What I had requested when we talked in our meeting was a `condition.name` column and a `condition.value` column. We would expect conditions to have different names and that callers would want to group by condition name. A bunch of separate columns for each different condition name makes that difficult.

When you update this don't forget to update the top-level function documentation.

nisystemlink/clients/spec/utilities/_dataframe_utilities.py
```diff
@@ -228,7 +270,7 @@ def __generate_condition_column_header(condition: Condition) -> str:
     return f"condition_{name}{unit}"
```
rbell517, line 270:
```suggestion
    return f"{DataFrameHeaders.CONDITION_COLUMN_HEADER_PREFIX}{name}{unit}"
```

rbell517 (approved):

nisystemlink/clients/spec/utilities/_dataframe_utilities.py
rbell517: I missed that this implementation is one row per spec, not one row per condition, sorry about that. You're right I had intended one row per condition to be the default behavior and you could then group by specId to get all the conditions for that specId. I suspect that kind of grouping is easier to work with because the columns are known labels rather needing some kind of condition label lookup to know the column labels, but I admit my pandas experience is limited. You're closer to the customer on this so I'll trust your judgement on what would be most common.

Since the implementation is simple I do think its worthwhile to go ahead and include out of the box formatters for both condition per column and condition per row but I won't hold up the review asking for that.

nisystemlink/clients/spec/utilities/_dataframe_utilities.py
rbell517: Or in light of the other comments, `condition_per_column`

nisystemlink/clients/spec/utilities/_dataframe_utilities.py
santhoshramaraj: It is best to deliver rows per spec and conditions as columns. I have seen every customer handle it that way because that is the intuitive way to represent the data.

nisystemlink/clients/spec/utilities/_dataframe_utilities.py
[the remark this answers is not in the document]
santhoshramaraj: In general, MultiIndex is an advanced topic in pandas. It is uncommon for users to encounter it initially. It is best to leave it unindexed and let the user decide what columns to merge by.

SSSantosh18 (approved):

nisystemlink/clients/spec/utilities/_dataframe_utilities.py
SSSantosh18: Thanks for your inputs team, condition per column has been made the default. condition per row formatter has also been included

SSadaiyappan-NI (approved):

PR #103, opened 2025-03-10
feat: Export steps as dataframe

ShriramS-NI:

- Adds functionality to return a normalized data frame, given a list of steps as input.
- Before the process of json normalization, inputs and outputs are restructured. From an array of names and values, they are converted into a dictionary with the name as the key, mapped to the value.
- Grouping of fields are done. This is much needed because, during normalization and the data frame conversion process, if there are new fields (especially fields which are nested) found in the successive steps, then there could be inconsistency in the ordering of the columns. (Also documented in the function's doc string)

This utility can be useful to get the steps data in the form of a Data frame which can be useful for further processing.

Unit tests are added

### DIFF
--- a/nisystemlink/clients/testmonitor/utilities/_dataframe_utilities.py
+++ b/nisystemlink/clients/testmonitor/utilities/_dataframe_utilities.py
@@ -1,4 +1,4 @@
-from typing import Any, Dict, List
+from typing import Any, Callable, Dict, List, Optional
 
 import pandas as pd
 from nisystemlink.clients.testmonitor.models import Result, Step, StepProjection
@@ -39,11 +39,19 @@ def convert_results_to_dataframe(
     return normalized_dataframe
 
 
-def convert_steps_to_dataframe(steps: List[Step]) -> pd.DataFrame:
+def convert_steps_to_dataframe(
+    steps: List[Step],
+    is_valid_measurement: Optional[Callable[[Dict[str, Any]], bool]] = None,
+) -> pd.DataFrame:
     """Converts a list of steps into a normalized dataframe.
 
     Args:
         steps: A list of steps.
+        is_valid_measurement: Optional function to check if a measurement is valid. The method takes
+            a dictionary as input and returns a boolean value. If the function is not provided, the
+            default behavior is to keep only those measurements that have both 'name' and 'measurement' fields.
+            If none of the measurement data have the desired fields, the data.parameters will not
+            appear in the dataframe.
 
     Returns:
         DataFrame:
@@ -57,7 +65,7 @@ def convert_steps_to_dataframe(steps: List[Step]) -> pd.DataFrame:
             all other step fields are duplicated.
     """
     DATA_PARAMETERS = "data.parameters"
-    step_dicts = __convert_steps_to_dict(steps)
+    step_dicts = __convert_steps_to_dict(steps, is_valid_measurement)
     steps_dataframe = pd.json_normalize(step_dicts, sep=".")
     steps_dataframe = __explode_and_normalize(
         steps_dataframe, DATA_PARAMETERS, f"{DATA_PARAMETERS}."
@@ -156,7 +164,10 @@ def __is_property_header(header: str) -> bool:
     return header.startswith(DataFrameHeaders.PROPERTY_COLUMN_HEADER_PREFIX)
 
 
-def __convert_steps_to_dict(steps: List[Step]) -> List[Dict[str, Any]]:
+def __convert_steps_to_dict(
+    steps: List[Step],
+    is_valid_measurement: Optional[Callable[[Dict[str, Any]], bool]] = None,
+) -> List[Dict[str, Any]]:
     """Converts a list of steps to dictionaries, excluding None values.
 
     Args:
@@ -168,10 +179,42 @@ def __convert_steps_to_dict(steps: List[Step]) -> List[Dict[str, Any]]:
     steps_dict = []
     for step in steps:
         single_step_dict = step.dict(exclude_none=True)
+
+        if step.data is not None and step.data.parameters is not None:
+            single_step_dict["data"]["parameters"] = __get_valid_data_parameters(
+                single_step_dict, is_valid_measurement
+            )
+
         __normalize_inputs_outputs(single_step_dict, step)
         __normalize_step_status(single_step_dict)
         steps_dict.append(single_step_dict)
     return steps_dict
+
+
+def __get_valid_data_parameters(
+    step_dict: Dict[str, Any],
+    is_valid_measurement: Optional[Callable[[Dict[str, Any]], bool]] = None,
+) -> List[Dict[str, Any]]:
+    """Gets valid measurement data parameters from the step dictionary.
+
+    Args:
+        step_dict: A dictionary with step information.
+        is_valid_measurement: Optional callback function to check if a measurement is valid. The method takes
+            a dictionary as input and returns a boolean value. If the function is not provided, the
+            default behavior is to keep only those measurements that have both 'name' and 'measurement' fields.
+
+    Returns:
+        List[Dict[str, Any]]: A list of dictionaries containing valid measurement data.
+    """
+    allowed_measurement_keys = ["name", "units"]
+    valid_measurement_parameters = []
+    for parameter in step_dict["data"]["parameters"]:
+        if (is_valid_measurement and is_valid_measurement(parameter)) or all(
+            key in parameter for key in allowed_measurement_keys
+        ):
+            valid_measurement_parameters.append(parameter)
+
+    return valid_measurement_parameters
 
 
 def __normalize_inputs_outputs(
--- a/nisystemlink/clients/testmonitor/utilities/_dataframe_utilities.py
+++ b/nisystemlink/clients/testmonitor/utilities/_dataframe_utilities.py
@@ -10,7 +10,7 @@ from nisystemlink.clients.testmonitor.models import (
 from nisystemlink.clients.testmonitor.utilities.constants import DataFrameHeaders
 
 
-def is_valid_measurement(measurement: Measurement) -> bool:
+def is_step_data_with_name_and_measurement(measurement: Measurement) -> bool:
     """Checks if a step data parameter is measurement data by ensuring it has both
     'name' and 'measurement' fields.
 
@@ -60,7 +60,7 @@ def convert_steps_to_dataframe(
     steps: List[Step],
     is_valid_measurement: Optional[
         Callable[[Measurement], bool]
-    ] = is_valid_measurement,
+    ] = is_step_data_with_name_and_measurement,
 ) -> pd.DataFrame:
     """Converts a list of steps into a normalized dataframe.
 
--- a/nisystemlink/clients/testmonitor/utilities/_dataframe_utilities.py
+++ b/nisystemlink/clients/testmonitor/utilities/_dataframe_utilities.py
@@ -1,22 +1,25 @@
 from typing import Any, Callable, Dict, List, Optional
 
 import pandas as pd
-from nisystemlink.clients.testmonitor.models import Result, Step, StepProjection
+from nisystemlink.clients.testmonitor.models import (
+    Measurement,
+    Result,
+    Step,
+    StepProjection,
+)
 from nisystemlink.clients.testmonitor.utilities.constants import DataFrameHeaders
 
 
-def is_step_data_with_name_and_measurement(data: Dict[str, Any]) -> bool:
-    """Checks if a step data parameter is a measurement data by ensuring it 
-    has both 'name' and 'measurement' fields.
+def is_step_data_with_name_and_measurement(measurement: Measurement) -> bool:
+    """Checks if a measurement data has both 'name' and 'measurement' fields.
 
     Args:
-        measurement: A dictionary containing measurement data.
+        measurement: A measurement data object
 
     Returns:
         bool: True if the measurement has both 'name' and 'measurement' fields, False otherwise.
     """
-    required_measurement_fields = ["name", "measurement"]
-    return set(required_measurement_fields).issubset(set(data.keys()))
+    return measurement.name is not None and measurement.measurement is not None
 
 
 def convert_results_to_dataframe(
@@ -55,7 +58,7 @@ def convert_results_to_dataframe(
 def convert_steps_to_dataframe(
     steps: List[Step],
     is_measurement_data_parameter: Optional[
-        Callable[[Dict[str, Any]], bool]
+        Callable[[Measurement], bool]
     ] = is_step_data_with_name_and_measurement,
 ) -> pd.DataFrame:
     """Converts a list of steps into a normalized dataframe.
@@ -64,8 +67,8 @@ def convert_steps_to_dataframe(
         steps: A list of steps.
         is_measurement_data_parameter: Optional callback function that checks if a step data parameter is a
             measurement so that only those are included in the returned dataframe. The method takes
-            a dictionary as input and returns a boolean value.
-            The default behavior is to consider only parameters that have both 'name' and 'measurement'
+            a measurement as input and returns a boolean value.
+            The default behavior is to consider only measurement data that have both 'name' and 'measurement'
             fields as measurement parameters.
             If none of the step data parameters have the required fields, the data.parameters will not
             appear in the dataframe.
@@ -184,23 +187,23 @@ def __is_property_header(header: str) -> bool:
 
 def __convert_steps_to_dict(
     steps: List[Step],
-    is_measurement_data_parameter: Optional[Callable[[Dict[str, Any]], bool]],
+    is_measurement_data_parameter: Optional[Callable[[Measurement], bool]],
 ) -> List[Dict[str, Any]]:
     """Converts a list of steps to dictionaries, excluding None values.
 
     Args:
         steps: A list of steps.
-        is_measurement_data_parameter: Optional callback function to check if a step
-            data parameter has the required fields.
+        is_measurement_data_parameter: Optional callback function that checks if a step data
+            parameter is a measurement so that only those are included in the returned dataframe.
 
     Returns:
         List[Dict[str, Any]]: A list of dictionaries containing step information.
     """
     steps_dict = []
     for step in steps:
-        single_step_dict = step.dict(exclude_none=True)
+        __filter_data_parameters(step, is_measurement_data_parameter)
 
-        __filter_data_parameters(single_step_dict, step, is_measurement_data_parameter)
+        single_step_dict = step.dict(exclude_none=True)
         __normalize_inputs_outputs(single_step_dict, step)
         __normalize_step_status(single_step_dict)
         steps_dict.append(single_step_dict)
@@ -208,35 +211,29 @@ def __convert_steps_to_dict(
 
 
 def __filter_data_parameters(
-    step_dict: Dict[str, Any],
     step: Step,
-    is_measurement_data_parameter: Optional[Callable[[Dict[str, Any]], bool]],
+    is_measurement_data_parameter: Optional[Callable[[Measurement], bool]],
 ) -> None:
     """Gets data parameters from the step dictionary and filters it based on the callback function.
 
     Args:
-        step_dict: A dictionary with step information.
         step: A Step object containing data parameters.
         is_measurement_data_parameter: Optional callback function to check if a measurement is valid. The method takes
             a dictionary as input and returns a boolean value. The default behavior is to consider only parameters
             that have both 'name' and 'measurement' fields as measurement parameters.
 
     Returns:
-        None: The function modifies step_dict in place with the filtered step data parameters.
+        None: The function modifies step parameters in place with filtered data parameters.
     """
-    valid_measurement_parameters = []
-    if (
-        is_measurement_data_parameter is not None
-        and step.data is not None
-        and step.data.parameters is not None
-    ):
-        for parameter in step_dict["data"]["parameters"]:
+    if step.data and step.data.parameters and is_measurement_data_parameter is not None:
+        valid_measurement_parameters = []
+        for measurement in step.data.parameters:
             if is_measurement_data_parameter and is_measurement_data_parameter(
-                parameter
+                measurement
             ):
-                valid_measurement_parameters.append(parameter)
+                valid_measurement_parameters.append(measurement)
 
-        step_dict["data"]["parameters"] = valid_measurement_parameters
+        step.data.parameters = valid_measurement_parameters
 
 
 def __normalize_inputs_outputs(
@@ -330,9 +327,7 @@ def __group_step_columns(dataframe_columns: List[str]) -> List[str]:
         StepProjection.DATA.lower(),
         StepProjection.PROPERTIES.lower(),
     ]
-    grouped_columns: Dict[str, List[str]] = {
-        category: [] for category in CATEGORY_KEYS
-    }
+    grouped_columns: Dict[str, List[str]] = {category: [] for category in CATEGORY_KEYS}
     for column in dataframe_columns:
         column_lower = column.lower()
         key = next(
--- a/nisystemlink/clients/testmonitor/utilities/_dataframe_utilities.py
+++ b/nisystemlink/clients/testmonitor/utilities/_dataframe_utilities.py
@@ -11,7 +11,8 @@ from nisystemlink.clients.testmonitor.utilities.constants import DataFrameHeader
 
 
 def is_step_data_with_name_and_measurement(measurement: Measurement) -> bool:
-    """Checks if a measurement data has both 'name' and 'measurement' fields.
+    """Checks if a step data parameter is measurement data by ensuring it has both
+    'name' and 'measurement' fields.
 
     Args:
         measurement: A measurement data object
--- a/nisystemlink/clients/testmonitor/utilities/_dataframe_utilities.py
+++ b/nisystemlink/clients/testmonitor/utilities/_dataframe_utilities.py
@@ -11,13 +11,6 @@ from pandas import DataFrame
 def convert_steps_to_dataframe(steps: List[Step]) -> DataFrame:
     """Converts a list of steps into a normalized dataframe.
 
-    - A new column would be created for unique `properties` across all steps. The property
-    columns would be named in the format `properties.property_name`.
-    - `Inputs` and `Outputs` are converted from a list of name-value pairs to a dict and then
-    normalized - similar to properties.
-    - For each `parameter` entry in `data`, a new row is added in the dataframe, with all the
-    other values are duplicated.
-
     Args:
         steps: A list of steps.
 
@@ -27,24 +20,23 @@ def convert_steps_to_dataframe(steps: List[Step]) -> DataFrame:
             fields in the input steps.
             - A new column would be created for unique `properties` across all steps. The property
             columns would be named in the format `properties.property_name`.
-            - `Inputs` and `Outputs` are converted from a list of name-value pairs to a dict and then
-            normalized - similar to properties.
-            - For each `parameter` entry in `data`, a new row is added in the dataframe, with all the
-            other values are duplicated.
+            - A new column would be created for unique `Inputs` and `Outputs` across all steps. The columns
+            would be named in the format `inputs.input_name` and `outputs.output_name` respectively.
+            - For each `parameter` entry in `data`, a new row is added in the dataframe, with data for
+            all other step fields are duplicated.
     """
     DATA_PARAMETERS = "data.parameters"
 
     restructured_steps = __restructure_steps(steps)
 
     steps_dataframe = pd.json_normalize(restructured_steps, sep=".")
-    if DATA_PARAMETERS in steps_dataframe.columns.to_list():
-        steps_dataframe = __explode_and_normalize(
-            steps_dataframe, DATA_PARAMETERS, f"{DATA_PARAMETERS}."
-        )
+    steps_dataframe = __explode_and_normalize(
+        steps_dataframe, DATA_PARAMETERS, f"{DATA_PARAMETERS}."
+    )
 
     grouped_columns = __group_step_columns(steps_dataframe.columns)
 
-    return steps_dataframe.reindex(columns=grouped_columns)
+    return steps_dataframe.reindex(columns=grouped_columns, copy=False)
 
 
 def __explode_and_normalize(
@@ -93,18 +85,18 @@ def __restructure_steps(steps: List[Step]) -> List[Dict[str, Any]]:
         List[Step]: Restructured steps - modification involves the conversion of list of inputs and outputs
         into dictionaries respectively.
     """
-    INPUTS = StepProjection.INPUTS.lower()
-    OUTPUTS = StepProjection.OUTPUTS.lower()
+    STEP_INPUTS = StepProjection.INPUTS.lower()
+    STEP_OUTPUTS = StepProjection.OUTPUTS.lower()
     restructured_steps = []
 
     for step in steps:
         step_dict = step.dict(exclude_none=True)
-        if INPUTS in step_dict:
-            step_dict[INPUTS] = (
+        if STEP_INPUTS in step_dict:
+            step_dict[STEP_INPUTS] = (
                 {item.name: item.value for item in step.inputs} if step.inputs else {}
             )
-        if OUTPUTS in step_dict:
-            step_dict[OUTPUTS] = (
+        if STEP_OUTPUTS in step_dict:
+            step_dict[STEP_OUTPUTS] = (
                 {item.name: item.value for item in step.outputs} if step.outputs else {}
             )
 
@@ -138,19 +130,16 @@ def __group_step_columns(dataframe_columns: List[str]) -> List[str]:
 
     for column in dataframe_columns:
         column_lower = column.lower()
-        if (
-            StepProjection.DATA.lower() in column_lower
-            and column != StepProjection.DATA_MODEL.lower()
-        ):
-            grouped_columns[StepProjection.DATA].append(column)
-        elif StepProjection.INPUTS.lower() in column_lower:
-            grouped_columns[StepProjection.INPUTS].append(column)
-        elif StepProjection.OUTPUTS.lower() in column_lower:
-            grouped_columns[StepProjection.OUTPUTS].append(column)
-        elif StepProjection.PROPERTIES.lower() in column_lower:
-            grouped_columns[StepProjection.PROPERTIES].append(column)
-        else:
-            grouped_columns[GENERAL_CATEGORIES].append(column)
+        key = next(
+            (
+                category
+                for category in CATEGORY_KEYS
+                if category.lower() in column_lower
+                and column != StepProjection.DATA_MODEL.lower()
+            ),
+            GENERAL_CATEGORIES,
+        )
+        grouped_columns[key].append(column)
 
     return [
         column for category in CATEGORY_KEYS for column in grouped_columns[category]
--- a/nisystemlink/clients/testmonitor/utilities/_dataframe_utilities.py
+++ b/nisystemlink/clients/testmonitor/utilities/_dataframe_utilities.py
@@ -1,0 +1,132 @@
+from typing import Any, Dict, List
+
+import pandas as pd
+from nisystemlink.clients.testmonitor.models import (
+    Step,
+    StepProjection,
+)
+from pandas import DataFrame
+
+
+def convert_steps_to_dataframe(steps: List[Step]) -> DataFrame:
+    """Converts a list of steps into a normalized dataframe.
+
+    - A new column would be created for unique `properties` across all steps. The property
+    columns would be named in the format `properties.property_name`.
+    - `Inputs` and `Outputs` are converted from a list of name-value pairs to a dict and then
+    normalized - similar to properties.
+    - For each `parameter` entry in `data`, a new row is added in the dataframe, with all the
+    other values are duplicated.
+
+    Args:
+        steps: A list of steps.
+
+    Returns:
+        DataFrame:
+            - A Pandas DataFrame containing the steps data. The DataFrame would consist of all the
+            fields in the input steps.
+            - A new column would be created for unique `properties` across all steps. The property
+            columns would be named in the format `properties.property_name`.
+            - `Inputs` and `Outputs` are converted from a list of name-value pairs to a dict and then
+            normalized - similar to properties.
+            - For each `parameter` entry in `data`, a new row is added in the dataframe, with all the
+            other values are duplicated.
+    """
+    DATA_PARAMETERS = "data.parameters"
+
+    restructured_steps = __restructure_steps(steps)
+
+    # checking if `data` exists in the steps. the following logic is specific to process `data` field
+    if steps and steps[0].data:
+        steps_dataframe = pd.json_normalize(restructured_steps, sep=".").explode(
+            DATA_PARAMETERS, ignore_index=True
+        )
+        steps_dataframe = pd.concat(
+            [
+                steps_dataframe.drop(columns=[DATA_PARAMETERS]),
+                pd.json_normalize(steps_dataframe[DATA_PARAMETERS]).add_prefix(
+                    f"{DATA_PARAMETERS}."
+                ),
+            ],
+            axis=1,
+        )
+    else:
+        steps_dataframe = pd.json_normalize(restructured_steps, sep=".")
+
+    grouped_columns = __group_step_columns(steps_dataframe.columns)
+
+    return steps_dataframe.reindex(columns=grouped_columns)
+
+
+def __restructure_steps(steps: List[Step]) -> List[Dict[str, Any]]:
+    """Restructures a list of step responses by converting input and output lists into dictionaries.
+
+    Each dictionary maps input/output names to their corresponding values, making it easier to
+    normalize the data into a DataFrame. Without this transformation, inputs and outputs would
+    remain as lists within a single cell.
+
+    Args:
+        steps: A list of step responses retrieved from the API.
+
+    Returns:
+        List[Step]: Restructured steps - modification involves the conversion of list of inputs and outputs
+        into dictionaries respectively.
+    """
+    restructured_steps = []
+
+    for step in steps:
+        step_dict = step.dict(exclude_none=True)
+        step_dict[StepProjection.INPUTS.lower()] = (
+            {item.name: item.value for item in step.inputs} if step.inputs else {}
+        )
+        step_dict[StepProjection.OUTPUTS.lower()] = (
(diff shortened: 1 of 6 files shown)


nisystemlink/clients/testmonitor/utilities/_dataframe_utilities.py
```diff
@@ -0,0 +1,157 @@
...
+
+def convert_steps_to_dataframe(steps: List[Step]) -> DataFrame:
+    """Converts a list of steps into a normalized dataframe.
+
+    - A new column would be created for unique `properties` across all steps. The property
+    columns would be named in the format `properties.property_name`.
+    - `Inputs` and `Outputs` are converted from a list of name-value pairs to a dict and then
+    normalized - similar to properties.
+    - For each `parameter` entry in `data`, a new row is added in the dataframe, with all the
+    other values are duplicated.
```
SSadaiyappan-NI, line 19:
```suggestion
    - For each `parameter` entry in `data`, a new row is added in the data frame, while all other column values will be duplicated.
```

nisystemlink/clients/testmonitor/utilities/_dataframe_utilities.py
```diff
@@ -0,0 +1,157 @@
...
+
+    steps_dataframe = pd.json_normalize(restructured_steps, sep=".")
+    if DATA_PARAMETERS in steps_dataframe.columns.to_list():
+        steps_dataframe = __explode_and_normalize(
+            steps_dataframe, DATA_PARAMETERS, f"{DATA_PARAMETERS}."
+        )
+
+    grouped_columns = __group_step_columns(steps_dataframe.columns)
+
+    return steps_dataframe.reindex(columns=grouped_columns)
```
SSadaiyappan-NI, line 47: can you check the other PR's to see how to optimize reindex

nisystemlink/clients/testmonitor/utilities/_dataframe_utilities.py
```diff
@@ -0,0 +1,157 @@
...
+            normalized - similar to properties.
+            - For each `parameter` entry in `data`, a new row is added in the dataframe, with all the
+            other values are duplicated.
+    """
+    DATA_PARAMETERS = "data.parameters"
+
+    restructured_steps = __restructure_steps(steps)
+
+    steps_dataframe = pd.json_normalize(restructured_steps, sep=".")
+    if DATA_PARAMETERS in steps_dataframe.columns.to_list():
```
SSadaiyappan-NI, line 40: we are checking if column in dataframe inside the explode method, do we need this  if check again?

nisystemlink/clients/testmonitor/utilities/_dataframe_utilities.py
```diff
@@ -0,0 +1,157 @@
...
+
+    Args:
+        steps: A list of step responses retrieved from the API.
+
+    Returns:
+        List[Step]: Restructured steps - modification involves the conversion of list of inputs and outputs
+        into dictionaries respectively.
+    """
+    INPUTS = StepProjection.INPUTS.lower()
+    OUTPUTS = StepProjection.OUTPUTS.lower()
```
SSadaiyappan-NI, line 97: make this as STEP_INPUTS and STEP_OUTPUTS

nisystemlink/clients/testmonitor/utilities/_dataframe_utilities.py
```diff
@@ -0,0 +1,157 @@
...
+    When normalizing steps into a dataframe, new input, output, or property fields may be added at the end,
+    disrupting the expected column order. This function ensures columns are grouped properly.
+
+    Args:
+        dataframe_columns: The list of all columns from the normalized dataframe.
+
+    Returns:
+        List[str]: A list containing grouped and ordered columns.
+    """
+    GENERAL_CATEGORIES = "general"
```
SSadaiyappan-NI, line 128: for grouping the columns together, can you check with Priya if she has explored other options, while creating dataframe or anyother options

nisystemlink/clients/testmonitor/utilities/_dataframe_utilities.py
```diff
@@ -0,0 +1,157 @@
...
+    CATEGORY_KEYS = [
```
SSadaiyappan-NI, line 129:
```
CATEGORY_KEYS = [
    GENERAL_CATEGORIES,
    StepProjection.INPUTS,
    StepProjection.OUTPUTS,
    StepProjection.DATA,
    StepProjection.PROPERTIES,
]

grouped_columns = {category: [] for category in CATEGORY_KEYS}

for column in dataframe_columns:
    column_lower = column.lower()
    key = next(
        (category for category in CATEGORY_KEYS if category.lower() in column_lower and column != StepProjection.DATA_MODEL.lower()), 
        GENERAL_CATEGORIES
    )
    grouped_columns[key].append(column)

return [col for category in CATEGORY_KEYS for col in grouped_columns[category]]
```
can you check this logic

nisystemlink/clients/testmonitor/utilities/_dataframe_utilities.py
SSSantosh18:
```suggestion
    - For each `parameter` entry in `data`, a new row is added in the dataframe, while data for all other result fields will be duplicated
```

nisystemlink/clients/testmonitor/utilities/_dataframe_utilities.py
```diff
@@ -0,0 +1,157 @@
...
+        normalized_dataframe = pd.json_normalize(
+            exploded_dataframe.pop(column)
+        ).add_prefix(prefix)
+
+        return pd.concat([exploded_dataframe, normalized_dataframe], axis=1)
+
+    return dataframe
+
+
+def __restructure_steps(steps: List[Step]) -> List[Dict[str, Any]]:
```
SSSantosh18, line 82:
```suggestion
def __normalize_inputs_and_outputs(steps: List[Step]) -> List[Dict[str, Any]]:
```

SSSantosh18: this method only acts on inputs and outputs. So, its not appropriate to have a generic name like restructure_steps

SSSantosh18: Looks like this method also converts this to dict, then this method should be named `__convert_steps_to_dict` and the input and output normalization logic should be moved to a new method with the name as I suggested above.

Its not good to make a method do too many things

nisystemlink/clients/testmonitor/utilities/_dataframe_utilities.py
```diff
@@ -0,0 +1,157 @@
...
+        StepProjection.DATA,
+        StepProjection.PROPERTIES,
+    ]
+
+    grouped_columns: Dict[str, List[str]] = {category: [] for category in CATEGORY_KEYS}
+
+    for column in dataframe_columns:
+        column_lower = column.lower()
+        if (
+            StepProjection.DATA.lower() in column_lower
```
SSSantosh18, line 142: What would happen if there is "data" in a property name, that case would cause problems, it would happen for all the below condition as well.

`Starts_with` would be a better operator

SSSantosh18: Also, I feel this logic is different from other df utilities. Can we maintain consistency and use similar logic or code?

tests/testmonitor/test_testmonitor_dataframe_utilities.py
```diff
@@ -0,0 +1,295 @@
...
+    StepData,
+)
+from nisystemlink.clients.testmonitor.utilities import convert_steps_to_dataframe
+from pandas import DataFrame
+
+
+@pytest.fixture
+def mock_steps_data() -> List[Step]:
+    """Fixture to return a mock step data."""
+    step1 = Step(
```
SSSantosh18, line 20: we can also have some steps where most of the fields is none

SSSantosh18: I see it is handled in the test. So, it is fine

tests/testmonitor/test_testmonitor_dataframe_utilities.py
```diff
@@ -0,0 +1,295 @@
...
+            NamedValue(name="Input22", value="input_value_22"),
+        ],
+        outputs=[
+            NamedValue(name="Output21", value="output_value_21"),
+            NamedValue(name="Output22", value="output_value_22"),
+        ],
+        data_model="STDF",
+        data=StepData(
+            text="data2",
+            parameters=[
```
SSSantosh18, line 92: 1. We can have some parameters with common names across steps
2. one step can have only one parameter

tests/testmonitor/test_testmonitor_dataframe_utilities.py
```diff
@@ -0,0 +1,295 @@
...
+        step_id="5ffb2bf6771fa11e877838dd6",
+        parent_id="5ffb2bf6771fa11e877838dd7",
+        result_id="5ffb2bf6771fa11e877838dd8",
+        path="Path",
+        path_ids=["path_id_21", "path_id_22"],
+        status=Status(status_type=StatusType.DONE, status_name="Done"),
+        total_time_in_seconds=5,
+        started_at=datetime(2023, 3, 27, 18, 39, 49, tzinfo=timezone.utc),
+        updated_at=datetime(2024, 2, 2, 14, 22, 4, 625255, tzinfo=timezone.utc),
+        inputs=[
```
SSSantosh18, line 81: 1. we can have some common inputs across the steps

nisystemlink/clients/testmonitor/utilities/_dataframe_utilities.py
priyadarshini-ni: The logic has been changed as suggested [here](https://github.com/ni/nisystemlink-clients-python/pull/103#discussion_r1987135466)

nisystemlink/clients/testmonitor/utilities/_dataframe_utilities.py
```diff
@@ -0,0 +1,157 @@
...
+    normalized - similar to properties.
+    - For each `parameter` entry in `data`, a new row is added in the dataframe, with all the
+    other values are duplicated.
+
+    Args:
+        steps: A list of steps.
+
+    Returns:
+        DataFrame:
+            - A Pandas DataFrame containing the steps data. The DataFrame would consist of all the
```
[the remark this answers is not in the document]
priyadarshini-ni: Moved the statements to returns

nisystemlink/clients/testmonitor/utilities/_dataframe_utilities.py
priyadarshini-ni: This works, updated to this

nisystemlink/clients/testmonitor/utilities/_dataframe_utilities.py
priyadarshini-ni: Added `copy=False` to avoid memory duplication

tests/testmonitor/test_testmonitor_dataframe_utilities.py
priyadarshini-ni: Added common input

nisystemlink/clients/testmonitor/utilities/_dataframe_utilities.py
priyadarshini-ni: Th `restructure_steps` method has been split into `__convert_steps_to_dict` and `__normalize_inputs_outputs`

ShriramS-NI:
```py
return [step.dict(exclude_none=True) for step in steps]
```

Here there is an iteration over all the steps. In `__normalize_inputs_outputs` again an iteration over all the steps is done.

```py
for step in steps_dict:
        if STEP_INPUTS in step:
```

Previously that's why we included the dict conversion and input, output restructuring in a single iteration. @priyadarshini-ni  @SSSantosh18 

```py
restructured_steps = []

    for step in steps:
        step_dict = step.dict(exclude_none=True)
        step_dict[StepProjection.INPUTS.lower()] = (
            {item.name: item.value for item in step.inputs} if step.inputs else {}
        )
        step_dict[StepProjection.OUTPUTS.lower()] = (
            {item.name: item.value for item in step.outputs} if step.outputs else {}
        )

        restructured_steps.append(step_dict)

    return restructured_steps
```

nisystemlink/clients/testmonitor/utilities/_dataframe_utilities.py
SSSantosh18: @priyadarshini-ni, @ShriramS-NI still the actual problem I mentioned is not addresssed

priyadarshini-ni: `startswith` has been used as suggested

nisystemlink/clients/testmonitor/utilities/_dataframe_utilities.py
```diff
@@ -1,57 +1,259 @@
...
+        set_id_as_index: If true, result id will be set as index for the dataframe.
+                         If false, index will not be set.
+                         Default value is false.
+
+    Returns:
+        A Pandas DataFrame with the each result fields having a separate column.
+        Following fields are split into sub-columns.
+            - status_type_summary: All the entries will be split into separate columns.
+            For example, status_type_summary.LOOPING, status_type_summary.PASSED, etc
+            - status: Split into status.status_type and status.status_name columns.
```
SSSantosh18, line 24:
```suggestion
```

SSSantosh18: Status is now modified into single column

nisystemlink/clients/testmonitor/utilities/_dataframe_utilities.py
```diff
@@ -39,6 +39,41 @@ def convert_results_to_dataframe(
     return normalized_dataframe
 
 
+def convert_steps_to_dataframe(
+    steps: List[Step],
+    is_valid_measurement: Optional[Callable[[Dict[str, Any]], bool]] = None,
```
SSSantosh18, line 44:
```suggestion
    is_measurement_data_parameter: Optional[Callable[[Dict[str, Any]], bool]] = None,
```

nisystemlink/clients/testmonitor/utilities/_dataframe_utilities.py
```diff
@@ -127,3 +162,155 @@ def __is_property_header(header: str) -> bool:
...
+                single_step_dict, is_valid_measurement
+            )
+
+        __normalize_inputs_outputs(single_step_dict, step)
+        __normalize_step_status(single_step_dict)
+        steps_dict.append(single_step_dict)
+    return steps_dict
+
+
+def __get_valid_data_parameters(
```
SSSantosh18, line 194:
```suggestion
def __get_measurement_data_parameters(
```

nisystemlink/clients/testmonitor/utilities/_dataframe_utilities.py
```diff
@@ -127,3 +162,155 @@ def __is_property_header(header: str) -> bool:
...
+        )
+    if STEP_OUTPUTS in step_dict:
+        step_dict[STEP_OUTPUTS] = (
+            {item.name: item.value for item in step.outputs} if step.outputs else {}
+        )
+
+
+def __normalize_step_status(step_dict: Dict[str, Any]) -> None:
+    step_status = step_dict.get("status", {})
+    if step_status.get("status_type") == "CUSTOM":
```
SSSantosh18, line 247:
```py
step_dict["status"] = step_status.get("status_name", None) if step_status.get("status_type") == "CUSTOM" else step_status["status_type"].value
```

SSSantosh18: can we use this?

nisystemlink/clients/testmonitor/utilities/_dataframe_utilities.py
```diff
@@ -127,3 +162,155 @@ def __is_property_header(header: str) -> bool:
...
+        StepProjection.DATA,
+        StepProjection.PROPERTIES,
+    ]
+    grouped_columns: Dict[str, List[str]] = {category: [] for category in CATEGORY_KEYS}
+    for column in dataframe_columns:
+        column_lower = column.lower()
+        key = next(
+            (
+                category
+                for category in CATEGORY_KEYS
```
SSSantosh18, line 307: category_keys also contains "general" but this will not match with any column but would be adding one more iteration for each column

can we remove that?

nisystemlink/clients/testmonitor/utilities/_dataframe_utilities.py
priyadarshini-ni: Updated

nisystemlink/clients/testmonitor/utilities/_dataframe_utilities.py
priyadarshini-ni: Removed

nisystemlink/clients/testmonitor/utilities/_dataframe_utilities.py
priyadarshini-ni: Removed "general" from the iteration

nisystemlink/clients/testmonitor/utilities/_dataframe_utilities.py
```diff
@@ -1,10 +1,23 @@
-from typing import Any, Dict, List
+from typing import Any, Callable, Dict, List, Optional
 
 import pandas as pd
-from nisystemlink.clients.testmonitor.models import Result
+from nisystemlink.clients.testmonitor.models import Result, Step, StepProjection
 from nisystemlink.clients.testmonitor.utilities.constants import DataFrameHeaders
 
 
+def is_default_measurement_data_parameter(measurement_data: Dict[str, Any]) -> bool:
```
SSSantosh18, line 8:
```suggestion
def is_step_data_with_name_and_measurement(data: Dict[str, Any]) -> bool:
```

nisystemlink/clients/testmonitor/utilities/_dataframe_utilities.py
```diff
@@ -1,10 +1,23 @@
...
+    """Checks if a measurement is valid by ensuring it has both 'name' and 'measurement' fields.
+
+    Args:
+        measurement: A dictionary containing measurement data.
+
+    Returns:
+        bool: True if the measurement has both 'name' and 'measurement' fields, False otherwise.
+    """
+    allowed_measurement_keys = ["name", "measurement"]
+    return all(key in measurement_data for key in allowed_measurement_keys)
```
SSSantosh18, line 18:
```py
set(required_measurement_fields).issubset(set(measurement_data.keys()))
```
is this a better option considering both readability and performance?

nisystemlink/clients/testmonitor/utilities/_dataframe_utilities.py
```diff
@@ -1,10 +1,23 @@
...
+def is_default_measurement_data_parameter(measurement_data: Dict[str, Any]) -> bool:
+    """Checks if a measurement is valid by ensuring it has both 'name' and 'measurement' fields.
+
+    Args:
+        measurement: A dictionary containing measurement data.
+
+    Returns:
+        bool: True if the measurement has both 'name' and 'measurement' fields, False otherwise.
+    """
+    allowed_measurement_keys = ["name", "measurement"]
```
SSSantosh18, line 17:
```suggestion
    required_measurement_fields = ["name", "measurement"]
```

nisystemlink/clients/testmonitor/utilities/_dataframe_utilities.py
```diff
@@ -39,6 +51,43 @@ def convert_results_to_dataframe(
...
+) -> pd.DataFrame:
+    """Converts a list of steps into a normalized dataframe.
+
+    Args:
+        steps: A list of steps.
+        is_measurement_data_parameter: Optional function to check if a measurement is valid. The method takes
+            a dictionary as input and returns a boolean value. If the function is not provided, the
+            default behavior is to keep only those measurements that have both 'name' and 'measurement' fields.
+            If none of the measurement data have the desired fields, the data.parameters will not
+            appear in the dataframe.
```
SSSantosh18, line 68: describe what would happen if the parameter is set as none

nisystemlink/clients/testmonitor/utilities/_dataframe_utilities.py
```diff
@@ -39,6 +51,43 @@ def convert_results_to_dataframe(
...
+    steps: List[Step],
+    is_measurement_data_parameter: Optional[
+        Callable[[Dict[str, Any]], bool]
+    ] = is_default_measurement_data_parameter,
+) -> pd.DataFrame:
+    """Converts a list of steps into a normalized dataframe.
+
+    Args:
+        steps: A list of steps.
+        is_measurement_data_parameter: Optional function to check if a measurement is valid. The method takes
```
SSSantosh18, line 64:
```suggestion
        is_measurement_data_parameter: Optional callback function that checks if a step data parameter is a measurement so that only the measurement data is included in the dataframe. The method takes
```

nisystemlink/clients/testmonitor/utilities/_dataframe_utilities.py
```diff
@@ -39,6 +51,43 @@ def convert_results_to_dataframe(
...
+            a dictionary as input and returns a boolean value. If the function is not provided, the
```
SSSantosh18, line 65:
```suggestion
            a dictionary as input and returns a boolean value. The
```

nisystemlink/clients/testmonitor/utilities/_dataframe_utilities.py
```diff
@@ -39,6 +51,43 @@ def convert_results_to_dataframe(
...
+            default behavior is to keep only those measurements that have both 'name' and 'measurement' fields.
```
SSSantosh18, line 66:
```suggestion
            default behavior is to consider only parameter that have both 'name' and 'measurement' fields as measurement parameters.
```

nisystemlink/clients/testmonitor/utilities/_dataframe_utilities.py
```diff
@@ -39,6 +51,43 @@ def convert_results_to_dataframe(
...
+            If none of the measurement data have the desired fields, the data.parameters will not
```
SSSantosh18, line 67:
```suggestion
            If none of the step data parameters data have the desired fields, the data.parameters will not
```

nisystemlink/clients/testmonitor/utilities/_dataframe_utilities.py
```diff
@@ -127,3 +176,157 @@ def __is_property_header(header: str) -> bool:
...
+        steps: A list of steps.
+
+    Returns:
+        List[Dict[str, Any]]: A list of dictionaries containing step information.
+    """
+    steps_dict = []
+    for step in steps:
+        single_step_dict = step.dict(exclude_none=True)
+
+        if (
```
SSSantosh18, line 197: can be moved to a separate method `__filter_data_parameters`

nisystemlink/clients/testmonitor/utilities/_dataframe_utilities.py
```diff
@@ -127,3 +176,157 @@ def __is_property_header(header: str) -> bool:
...
+        StepProjection.PROPERTIES,
+    ]
+    grouped_columns: Dict[str, List[str]] = {category: [] for category in CATEGORY_KEYS}
+    for column in dataframe_columns:
+        column_lower = column.lower()
+        key = next(
+            (
+                category
+                for category in CATEGORY_KEYS[1:]
+                if column_lower.startswith(category.lower())
```
SSSantosh18, line 324: if we can convert the case to lower in the source list itself, there would be no need to lower case for each iteration of the parent loop

tests/testmonitor/test_testmonitor_dataframe_utilities.py
```diff
@@ -161,6 +280,112 @@ def test__convert_results_to_dataframe_with_no_results__returns_empty_dataframe(
...
+        desired_measurement_keys = ["name", "temperature"]
+        expected_steps_dataframe = self.__get_expected_steps_dataframe(
+            mock_steps_data, desired_measurement_keys
+        )
+
+        steps_dataframe = convert_steps_to_dataframe(
+            mock_steps_data, is_measurement_data_parameter
+        )
+
+        # import pdb
```
SSSantosh18, line 372: remove commented code

tests/testmonitor/test_testmonitor_dataframe_utilities.py
```diff
@@ -161,6 +280,112 @@ def test__convert_results_to_dataframe_with_no_results__returns_empty_dataframe(
...
+        assert not steps_dataframe.empty
+        assert (
+            steps_dataframe.columns.to_list()
+            == expected_steps_dataframe.columns.to_list()
+        )
+        pd.testing.assert_frame_equal(
+            steps_dataframe, expected_steps_dataframe, check_dtype=True
+        )
+
+    def test__convert_steps_to_dataframe_with_is_callback__returns_dataframe_with_valid_measurement(
```
SSSantosh18, line 355:
```suggestion
    def test__convert_steps_to_dataframe_with_callback__returns_dataframe_with_valid_measurement(
```

SSSantosh18 (commented):
@rbell517, there are some comments to fix but not major, so requesting your review to avoid delays

nisystemlink/clients/testmonitor/utilities/_dataframe_utilities.py
priyadarshini-ni: Added behavior when parameter is set to None

nisystemlink/clients/testmonitor/utilities/_dataframe_utilities.py
priyadarshini-ni: Used a source list with lower case

nisystemlink/clients/testmonitor/utilities/_dataframe_utilities.py
priyadarshini-ni: Updated as suggested

nisystemlink/clients/testmonitor/utilities/_dataframe_utilities.py
priyadarshini-ni: Modified as `Optional callback function that checks if a step data parameter is a required measurement so that only those are included in the resultant dataframe`

nisystemlink/clients/testmonitor/utilities/_dataframe_utilities.py
```diff
@@ -127,3 +179,174 @@ def __is_property_header(header: str) -> bool:
...
+    """
+    GENERAL_CATEGORIES = "general"
+    CATEGORY_KEYS = [
+        GENERAL_CATEGORIES,
+        StepProjection.INPUTS,
+        StepProjection.OUTPUTS,
+        StepProjection.DATA,
+        StepProjection.PROPERTIES,
+    ]
+    category_keys_lower = [category.lower() for category in CATEGORY_KEYS]
```
SSSantosh18, line 332: I don't see any reason to maintain a separate list of lower strings, the source itself could be created with lower version of these strings

SSSantosh18 (approved):

SSSantosh18 (approved):

SSSantosh18:
> @rbell517, there are some comments to fix but not major, so requesting your review to avoid delays

Most of the comments are resolved and ready for review

nisystemlink/clients/testmonitor/utilities/_dataframe_utilities.py
```diff
@@ -55,17 +55,20 @@ def convert_steps_to_dataframe(
...
     """Converts a list of steps into a normalized dataframe.
 
     Args:
         steps: A list of steps.
-        is_measurement_data_parameter: Optional function to check if a measurement is valid. The method takes
-            a dictionary as input and returns a boolean value. If the function is not provided, the
-            default behavior is to keep only those measurements that have both 'name' and 'measurement' fields.
-            If none of the measurement data have the desired fields, the data.parameters will not
+        is_measurement_data_parameter: Optional callback function that checks if a step data parameter is a
+            required measurement so that only those are included in the resultant dataframe. The method takes
```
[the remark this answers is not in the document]
rbell517:
```suggestion
            measurement so that only those are included in the returned dataframe. The method takes
```

nisystemlink/clients/testmonitor/utilities/_dataframe_utilities.py
```diff
@@ -127,3 +179,174 @@ def __is_property_header(header: str) -> bool:
...
+        and step.data is not None
+        and step.data.parameters is not None
+    ):
+        for parameter in step_dict["data"]["parameters"]:
+            if is_measurement_data_parameter and is_measurement_data_parameter(
+                parameter
+            ):
+                valid_measurement_parameters.append(parameter)
+
+        step_dict["data"]["parameters"] = valid_measurement_parameters
```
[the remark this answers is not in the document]
SSSantosh18: @rbell517, I agree there will be noise if we purely think from the measurement perspective and I think getting the steps with measurement would be primary use case if you are trying to export the data outside SystemLink.

But I feel its better to keep the utility open to fetch non measurement steps as well so it could be useful in example cases you mentioned like analysing Teststand step failure. I am thinking by default we can provide all the steps in the dataframe and include one optional parameter (bool flag) to say take only measurement steps in future if we think there will be more customers who will have such non-measurement steps.

@santhoshramaraj, please add your thoughts as well

nisystemlink/clients/testmonitor/utilities/_dataframe_utilities.py
```diff
@@ -39,6 +51,46 @@ def convert_results_to_dataframe(
...
+            - A new column would be created for unique `Inputs` and `Outputs` across all steps. The columns
+            would be named in the format `inputs.input_name` and `outputs.output_name` respectively.
+            - For each `parameter` entry in `data`, a new row is added in the dataframe, with data for
+            all other step fields are duplicated.
+    """
+    DATA_PARAMETERS = "data.parameters"
+    step_dicts = __convert_steps_to_dict(steps, is_measurement_data_parameter)
+    steps_dataframe = pd.json_normalize(step_dicts, sep=".")
+    steps_dataframe = __explode_and_normalize(
+        steps_dataframe, DATA_PARAMETERS, f"{DATA_PARAMETERS}."
```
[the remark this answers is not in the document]
SSSantosh18: I agree this adds more clarity, `data.measurement.name` would help identify later which step field it was part of.

This utility also allows non-measurement data to be included if the callback is set as none, will `data.non_measurement.<field_name>` work in that case?

@santhoshramaraj, add your thoughts as well

nisystemlink/clients/testmonitor/utilities/_dataframe_utilities.py
priyadarshini-ni: Changed to lower in the source itself

nisystemlink/clients/testmonitor/utilities/_dataframe_utilities.py
```diff
@@ -39,6 +51,46 @@ def convert_results_to_dataframe(
     return normalized_dataframe
 
 
+def convert_steps_to_dataframe(
+    steps: List[Step],
+    is_measurement_data_parameter: Optional[
+        Callable[[Dict[str, Any]], bool]
```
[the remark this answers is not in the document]
priyadarshini-ni: The callable function has been changed to get a `Measurement` arg. The code flow has also been updated for the same.

tests/testmonitor/test_testmonitor_dataframe_utilities.py
```diff
@@ -75,7 +82,119 @@ def results() -> List[Result]:
...
+                    "comparisonType": "GTLT",
+                },
+                {
+                    "name": "parameter_22",
+                    "units": "C",
+                    "status": "Passed",
+                    "lowLimit": "7.0",
+                    "highLimit": "22.0",
+                    "comparisonType": "GTLT",
+                },
```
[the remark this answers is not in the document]
priyadarshini-ni: Added test case to verify
1. Step with data parameter keys other than the measurement fields like "additional_data"
2. Step with a single data parameter containing just "name" and "status".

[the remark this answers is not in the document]
priyadarshini-ni: Added extra fields other than the measurement type fields

nisystemlink/clients/testmonitor/utilities/_dataframe_utilities.py
```diff
@@ -1,10 +1,23 @@
...
+    """Checks if a measurement is required by ensuring it has both 'name' and 'measurement' fields.
+
+    Args:
+        measurement: A dictionary containing measurement data.
+
+    Returns:
+        bool: True if the measurement has both 'name' and 'measurement' fields, False otherwise.
+    """
+    required_measurement_fields = ["name", "measurement"]
+    return set(required_measurement_fields).issubset(set(data.keys()))
```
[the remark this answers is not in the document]
priyadarshini-ni: Updated as suggested

tests/testmonitor/test_testmonitor_dataframe_utilities.py
```diff
@@ -215,3 +438,114 @@ def __get_expected_results_dataframe(self, results: List[Result]):
...
+                else {}
+            )
+
+            restructured_step.update(
+                {f"outputs.{item.name}": item.value for item in step.outputs}
+                if step.outputs
+                else {}
+            )
+
+            if step.data and step.data.parameters:
```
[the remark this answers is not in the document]
priyadarshini-ni: Simplified the expected dataframe function. The expected data parameters dict has been passed from the test case itself.

tests/testmonitor/test_testmonitor_dataframe_utilities.py
```diff
@@ -215,3 +612,104 @@ def __get_expected_results_dataframe(self, results: List[Result]):
...
 
         return results_df
+
+    def __get_expected_steps_dataframe(
+        self,
+        mock_steps_data: List[Step],
+        data_parameter_columns=None,
+        expected_column_order=None,
+    ) -> pd.DataFrame:
+        restructured_mock_steps = []
```
SSSantosh18, line 622:
```suggestion
        steps_with_measurement_per_row = []
```

tests/testmonitor/test_testmonitor_dataframe_utilities.py
```diff
@@ -384,6 +421,143 @@ def is_measurement_data_parameter(step_dict: Dict[str, Any]) -> bool:
...
+                2024, 2, 2, 14, 22, 4, 625255, tzinfo=datetime.timezone.utc
+            ),
+            inputs=[
+                NamedValue(name="Input11", value="input_value_123"),
+            ],
+            outputs=[
+                NamedValue(name="Output11", value="output_value_123"),
+            ],
+            data_model="STDF",
+            data=StepData(
```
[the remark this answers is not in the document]
priyadarshini-ni: The other fields have been removed

tests/testmonitor/test_testmonitor_dataframe_utilities.py
```diff
@@ -215,3 +612,104 @@ def __get_expected_results_dataframe(self, results: List[Result]):
...
+                    "started_at": step.started_at,
+                    "updated_at": step.updated_at,
+                    "data_model": step.data_model,
+                    "has_children": step.has_children,
+                    "workspace": step.workspace,
+                    "keywords": step.keywords,
+                    "data.text": step.data.text if step.data else None,
+                    **{key: value for key, value in (parameter.items() or {})},
+                }
+                restructured_step.update(
```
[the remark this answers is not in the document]
priyadarshini-ni: Updated as suggested

nisystemlink/clients/testmonitor/utilities/_dataframe_utilities.py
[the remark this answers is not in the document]
rbell517: If None is passed for the is_measurement filter the original `data.parameters` you started with might be better since we don't know if these are measurements or not. I don't love having these differences in output, but I think its better to bias toward the expected pattern.

nisystemlink/clients/testmonitor/utilities/_dataframe_utilities.py
```diff
@@ -1,21 +1,26 @@
...
+    Result,
+    Step,
+    StepProjection,
+)
 from nisystemlink.clients.testmonitor.utilities.constants import DataFrameHeaders
 
 
-def is_step_data_with_name_and_measurement(data: Dict[str, Any]) -> bool:
-    """Checks if a measurement is required by ensuring it has both 'name' and 'measurement' fields.
+def is_step_data_with_name_and_measurement(measurement: Measurement) -> bool:
```
rbell517, line 13: Lets simplify the name now that this operates on a Measurement
```suggestion
def has_name_and_measurement(measurement: Measurement) -> bool:
```

nisystemlink/clients/testmonitor/utilities/_dataframe_utilities.py
```diff
@@ -183,59 +188,51 @@ def __is_property_header(header: str) -> bool:
...
-            ):
-                valid_measurement_parameters.append(parameter)
+    if step.data and step.data.parameters and is_valid_measurement is not None:
+        valid_measurement_parameters = []
+        for measurement in step.data.parameters:
+            if is_valid_measurement and is_valid_measurement(measurement):
+                valid_measurement_parameters.append(measurement)
 
-        step_dict["data"]["parameters"] = valid_measurement_parameters
+        step.data.parameters = valid_measurement_parameters
```
rbell517, line 235: This is going to modify the Step that was input to the conversion, and we don't want that. Those inputs should be left unchanged.

nisystemlink/clients/testmonitor/utilities/_dataframe_utilities.py
```diff
@@ -324,21 +321,18 @@ def __group_step_columns(dataframe_columns: List[str]) -> List[str]:
...
     CATEGORY_KEYS = [
         GENERAL_CATEGORIES,
-        StepProjection.INPUTS,
-        StepProjection.OUTPUTS,
-        StepProjection.DATA,
-        StepProjection.PROPERTIES,
+        StepProjection.INPUTS.lower(),
+        StepProjection.OUTPUTS.lower(),
+        StepProjection.DATA.lower(),
+        StepProjection.PROPERTIES.lower(),
```
rbell517, line 327: Lower-casing these in each call to this function is wasteful if you're operating a large number of steps. Either drop using StepProjection and just use a hard coded string or make these once.

rbell517 (approved):

nisystemlink/clients/testmonitor/utilities/_dataframe_utilities.py
priyadarshini-ni: Used hardcoded lower case string instead

nisystemlink/clients/testmonitor/utilities/_dataframe_utilities.py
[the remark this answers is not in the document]
priyadarshini-ni: The prefix has been changed as suggested. `data.paramters` if the is_valid_measurement is None else, `data.measurement` is assigned.

nisystemlink/clients/testmonitor/utilities/_dataframe_utilities.py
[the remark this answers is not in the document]
SSSantosh18: We are going with including steps without measurement also for now. We can include the optional flag I was talking if needed in future

nisystemlink/clients/testmonitor/utilities/_dataframe_utilities.py
```diff
@@ -81,11 +86,13 @@ def convert_steps_to_dataframe(
             - For each `parameter` entry in `data`, a new row is added in the dataframe, with data for
             all other step fields are duplicated.
```
SSSantosh18, line 87: there should be a documentation on the change in column name for parameters based on callback function none or not

nisystemlink/clients/testmonitor/utilities/_dataframe_utilities.py
```diff
@@ -183,59 +190,55 @@ def __is_property_header(header: str) -> bool:
...
+    is_valid_measurement: Optional[Callable[[Measurement], bool]],
 ) -> None:
     """Gets data parameters from the step dictionary and filters it based on the callback function.
 
     Args:
         step_dict: A dictionary with step information.
         step: A Step object containing data parameters.
-        is_measurement_data_parameter: Optional callback function to check if a measurement is valid. The method takes
+        is_valid_measurement: Optional callback function to check if a measurement is valid. The method takes
             a dictionary as input and returns a boolean value. The default behavior is to consider only parameters
```
SSSantosh18, line 227:
```suggestion
            a Measurement as input and returns a boolean value. The default behavior is to consider only parameters
```

nisystemlink/clients/testmonitor/utilities/_dataframe_utilities.py
```diff
@@ -127,3 +186,167 @@ def __is_property_header(header: str) -> bool:
...
+        - A new DataFrame with the exploded rows and the normalized columns, all combined
+        with the original data in the dataframe.
+        - If the column is not found in the dataframe, the original dataframe is returned unchanged.
+    """
+    if column in dataframe:
+        exploded_dataframe = dataframe.explode(column, ignore_index=True)
+        normalized_dataframe = pd.json_normalize(
+            exploded_dataframe.pop(column)
+        ).add_prefix(prefix)
+        return pd.concat([exploded_dataframe, normalized_dataframe], axis=1)
```
SSSantosh18, line 312: Please use `copy=False`

SSSantosh18 (approved):

nisystemlink/clients/testmonitor/utilities/_dataframe_utilities.py
```diff
@@ -324,21 +327,18 @@ def __group_step_columns(dataframe_columns: List[str]) -> List[str]:
     GENERAL_CATEGORIES = "general"
     CATEGORY_KEYS = [
         GENERAL_CATEGORIES,
-        StepProjection.INPUTS,
-        StepProjection.OUTPUTS,
-        StepProjection.DATA,
-        StepProjection.PROPERTIES,
+        "inputs",
```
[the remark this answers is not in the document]
SSSantosh18: image

this also can be moved here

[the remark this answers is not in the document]
priyadarshini-ni: Moved the category headers to constants.py file

nisystemlink/clients/testmonitor/utilities/_dataframe_utilities.py
priyadarshini-ni: Updated

PR #104, opened 2025-03-12
fix: Delete product created during result client test fixture

priyadarshini-ni:

In the Result Client test cases, a [fixture](https://github.com/ni/nisystemlink-clients-python/blob/820df0e39bea16dfe9f8bc5fad21a57efbb0194b/tests/integration/testmonitor/test_testmonitor.py#L49) is used to create results with a dummy part number. As per the design of the create-results API in SLE, a product is created with that dummy part number. This product should be deleted in the fixture itself once the linked results are deleted.

Implementation - All the products created during a test case are deleted by the fixture.

As the test suites are executed in the pipeline, the tests are continuously run which ended up polluting the test environment.

Verified the existing cases are passing

### DIFF
--- a/tests/integration/testmonitor/test_testmonitor.py
+++ b/tests/integration/testmonitor/test_testmonitor.py
@@ -6,8 +6,8 @@ from nisystemlink.clients.core._api_exception import ApiException
 from nisystemlink.clients.core._http_configuration import HttpConfiguration
 from nisystemlink.clients.product._product_client import ProductClient
 from nisystemlink.clients.product.models._query_products_request import (
-    ProductProjection,
-    QueryProductsRequest,
+    ProductField,
+    QueryProductValuesRequest
 )
 from nisystemlink.clients.testmonitor import TestMonitorClient
 from nisystemlink.clients.testmonitor.models import (
@@ -45,7 +45,7 @@ def client(enterprise_config: HttpConfiguration) -> TestMonitorClient:
 
 @pytest.fixture(scope="class")
 def product_client(enterprise_config: HttpConfiguration) -> ProductClient:
-    """Fixture to create a TestMonitorClient instance."""
+    """Fixture to create a ProductClient instance."""
     return ProductClient(enterprise_config)
 
 
@@ -75,20 +75,18 @@ def create_results(client: TestMonitorClient, product_client: ProductClient):
         if response.results:
             created_results = created_results + response.results
 
-    part_number_of_created_product = created_results[0].part_number
+    part_numbers_of_created_products = list(set(result.part_number for result in created_results))
 
     client.delete_results(ids=[str(result.id) for result in created_results])
 
-    product = product_client.query_products_paged(
-        QueryProductsRequest(
-            filter=f'partNumber="{part_number_of_created_product}"',
-            projection=[ProductProjection.ID],
-            take=1,
+    filter_string = " or ".join([f'partNumber="{part_number}"' for part_number in part_numbers_of_created_products])
+    created_product_ids = product_client.query_product_values(
+        QueryProductValuesRequest(
+            filter=filter_string, field=ProductField.ID
         )
     )
 
-    if product.products[0].id:
-        product_client.delete_product(id=product.products[0].id)
+    product_client.delete_products(ids=created_product_ids)
 
 
 @pytest.fixture
@@ -161,16 +159,16 @@ class TestTestMonitor:
         assert created_result.serial_number == serial_number
 
     def test__create_multiple_results__multiple_creates_succeed(
-        self, client: TestMonitorClient, create_results, unique_identifier
+        self, client: TestMonitorClient, create_results
     ):
         program_name = "Test Program"
         status = Status.PASSED()
         results = [
             CreateResultRequest(
-                part_number=unique_identifier, program_name=program_name, status=status
+                part_number=uuid.uuid1().hex, program_name=program_name, status=status
             ),
             CreateResultRequest(
-                part_number=unique_identifier, program_name=program_name, status=status
+                part_number=uuid.uuid1().hex, program_name=program_name, status=status
             ),
         ]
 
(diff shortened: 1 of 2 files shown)


tests/integration/testmonitor/test_testmonitor.py
```diff
@@ -63,8 +74,22 @@ def _create_results(
     for response in responses:
         if response.results:
             created_results = created_results + response.results
+
+    part_number_of_created_product = created_results[0].part_number
+
     client.delete_results(ids=[str(result.id) for result in created_results])
 
+    product = product_client.query_products_paged(
```
SSSantosh18, line 82: Query product values would be the best fit for this use case as we are querying just a particular field

tests/integration/testmonitor/test_testmonitor.py
```diff
@@ -38,6 +43,12 @@ def client(enterprise_config: HttpConfiguration) -> TestMonitorClient:
     return TestMonitorClient(enterprise_config)
 
 
+@pytest.fixture(scope="class")
+def product_client(enterprise_config: HttpConfiguration) -> ProductClient:
+    """Fixture to create a TestMonitorClient instance."""
```
SSSantosh18, line 48:
```suggestion
    """Fixture to create a ProductClient instance."""
```

tests/integration/testmonitor/test_testmonitor.py
```diff
@@ -63,8 +74,22 @@ def _create_results(
     for response in responses:
         if response.results:
             created_results = created_results + response.results
+
+    part_number_of_created_product = created_results[0].part_number
```
SSSantosh18, line 78: await: multiple results can be created as part of this fixture. We should take all the partnumbers rather than just the first one

SSSantosh18 (changes requested):

tests/integration/testmonitor/test_testmonitor.py
priyadarshini-ni: Modified to use `query_product_values` API

tests/integration/testmonitor/test_testmonitor.py
priyadarshini-ni: Added functionality to delete multiple products

tests/integration/testmonitor/test_testmonitor.py
```diff
@@ -63,8 +74,25 @@ def _create_results(
     for response in responses:
         if response.results:
             created_results = created_results + response.results
+
+    part_numbers_of_created_products = list(
```
SSSantosh18, line 78: is it necessary to convert set as list?

tests/integration/testmonitor/test_testmonitor.py
```diff
@@ -63,8 +74,25 @@ def _create_results(
...
+        set(result.part_number for result in created_results)
+    )
+
     client.delete_results(ids=[str(result.id) for result in created_results])
 
+    filter_string = " or ".join(
```
SSSantosh18, line 84:
```suggestion
    product_filter_string = " or ".join(
```

SSSantosh18 (approved):

tests/integration/testmonitor/test_testmonitor.py
priyadarshini-ni: Removed the list conversion

tests/integration/testmonitor/test_testmonitor.py
```diff
@@ -63,8 +74,25 @@ def _create_results(
     for response in responses:
         if response.results:
             created_results = created_results + response.results
+
+    part_numbers_of_created_products = set(
```
rbell517, line 78: This cleanup implementation wouldn't change but we could reduce the work we're doing by not making unique part numbers. We only have a couple of test cases where we need a unique part of the result to query by, and logically that would be better to use serial number for. Otherwise I'd prefer we give it a part number that we can use to more easily see that these products come from our python client integration tests, like "Python Client Testing". And while we're at it, we should make a similar change for the result program name to make it more identifiable with this testing.

Here are the updated query tests with some other improvements like verifying projection behavior.
```python
def test__query_result_by_part_number_and_serial_number_with_projection__matches_expected_with_only_projected_fields(
        self, client: TestMonitorClient, create_results, unique_identifier
    ):
        results = [
            CreateResultRequest(
                part_number="Python Client Testing",
                program_name="Python Client Test Program",
                serial_number=unique_identifier,
                status=Status.PASSED()
            )
        ]
        create_response: CreateResultsPartialSuccess = create_results(results)
        assert create_response is not None
        
        query_request = QueryResultsRequest(
            filter=f'partNumber="{results[0].part_number}" and serialNumber="{results[0].serial_number}"',
            projection=[ResultProjection.ID, ResultProjection.PART_NUMBER, ResultProjection.SERIAL_NUMBER],
            return_count=True
        )
        query_response: PagedResults = client.query_results(query_request)
        
        assert query_response.total_count == 1
        assert len(query_response.results) == 1
        assert query_response.results[0].part_number == results[0].part_number
        assert query_response.results[0].serial_number == results[0].serial_number
        assert query_response.results[0].__fields_set__ == ["id", "part_number", "serial_number"]

def test__query_result_values_for_progam_name__name_matches(
        self, client: TestMonitorClient, create_results, unique_identifier
    ):
        results = [
            CreateResultRequest(
                part_number="Python Client Testing",
                program_name="Python Client Test Program",
                serial_number=unique_identifier,
                status=Status.PASSED()
            )
        ]
        create_response: CreateResultsPartialSuccess = create_results(results)
        assert create_response is not None
        
        query_request = QueryResultValuesRequest(
            filter=f'partNumber="{results[0].part_number}" and serialNumber="{results[0].serial_number}"', field=ResultField.PROGRAM_NAME
        )
        query_response: List[str] = client.query_result_values(query_request)
        
        assert query_response is not None
        assert len(query_response) == 1
        assert query_response[0] == results[0].program_name
```

rbell517 (approved):

tests/integration/testmonitor/test_testmonitor.py
priyadarshini-ni: The suggestions have been applied. Summary of changes made are listed below
1. Changing the part_number and program_name to "Python Client Testing" and "Python Client Test Program" resp for all results created
2. Using a combination of part_number and serial_number for query by test cases
3. No change made to products cleanup logic

PR #129, opened 2025-04-10
feat: Upgrade to Pydantic 2.11

### LINKED ISSUE #94: Support for Pydantic v2
Please provide support for Pydantic v2.
This prevents us from using this library in our environments, where we upgraded a year ago.

rbell517:

- Upgrades the version of Pydantic to v2.11.3
- Resolves #94 
- Greatly improves query performance with large takes due to validation time improvements in pydantic v2.11

The current query performance of the client is hampered by the pydantic v1.10 validation time. In addition being on an old version of pydantic limits some clients ability to use the library.

Existing unit and integration tests are passing. Manually tested the few APIs with changes that are not fully covered by our integration test coverage (notebooks, specifically)

To test query performance I used a notebook that retrieved test steps in 100,000 batches. The new query time is now roughly on par with using `requests` directly.


nisystemlink/clients/testmonitor/models/_step.py
```diff
@@ -32,7 +32,7 @@ class Step(JsonModel):
     status: Optional[Status] = None
     """The status of the step."""
 
-    total_time_in_seconds: Optional[int] = None
+    total_time_in_seconds: Optional[float] = None
```
adamarnesen, line 35: Is this change from `int` to `float` suggested by pydantic? This seems like it _may_ be an API break?

adamarnesen (approved):

nisystemlink/clients/testmonitor/models/_step.py
rbell517: Good question, this is actually a result of better pydantic 2 type strictness. It should be float and we use a float in the test, but it was silently being coerced to int previously. And `result` properly uses float for its version of the field. The integration tests caught this, implicitly.

As for API breaking, I'm inclined to not consider this breaking. From an runtime persepctive basically everything you can do with an int you can do with a float. The only exceptions are cases like array indexing, which make no sense for this field. The only real break would be if there is code out there that also uses int as the type hint and uses this field to fill that case. Mypy would flag that. I think thats OK, since that is dev-time only, so it would be safe to use the newer version at runtime.

adamarnesen: Sounds good to me. I don't have any other concerns with this one.

cameronwaterman (approved):

PR #130, opened 2025-04-18
fix: Prevent string discrete values which are string of numbers from passing as numeric condition values.

RSam-NI:

- It makes sure that string discrete values are correctly passing as string condition values numeric discrete values are correctly passing as numeric condition values.
- Data type of `discrete` value of `NumericConditionValue` is set to `List[Union[StringFloat, StrictInt]]`. Used StrictFloat and StrictInt because we only need integer and float values to be stored here and not string of numbers. If `int`/`float` are used, pydantic tries to convert the string of numbers into float/int.
- Data type of `discrete` value of `StringConditionValue` is set to `StrictStr`. Used `StrictStr` beause, we only want strings to be stored here and not any other datatypes. If `str` is used, pydantic converts any other datatypes to string and stores it.

- String discrete values which has string of numbers are being converted into numeric condition values.
- This PR has changes which converts such discrete values correctly as string condition values. And also makes sure that real numeric discrete values are being converted into numeric condition values.

### DIFF
--- a/nisystemlink/clients/spec/models/_condition.py
+++ b/nisystemlink/clients/spec/models/_condition.py
@@ -2,7 +2,7 @@ from enum import Enum
 from typing import List, Optional, Union
 
 from nisystemlink.clients.core._uplink._json_model import JsonModel
-from pydantic import StrictStr
+from pydantic import StrictFloat, StrictInt, StrictStr
 
 
 class ConditionType(Enum):
@@ -44,7 +44,10 @@ class NumericConditionValue(ConditionValueBase):
     range: Optional[List[ConditionRange]] = None
     """List of condition range values."""
 
-    discrete: Optional[List[float]] = None
+    # StrictFloat and StrictInt are used here because discrete is a common property for
+    # NumericConditionValue and StringConditionValue.
+    # And pydantic converts string of number into float/int by default if just float/int is used
+    discrete: Optional[List[Union[StrictFloat, StrictInt]]] = None
     """List of condition discrete values."""
 
     unit: Optional[str] = None
@@ -57,6 +60,9 @@ class StringConditionValue(ConditionValueBase):
     String conditions may only contain discrete lists of values.
     """
 
+    # StrictStr is used here because discrete is a common property for
+    # NumericConditionValue and StringConditionValue.
+    # And pydantic converts any datatype into string by default if just str is used.
     discrete: Optional[List[StrictStr]] = None
     """List of condition discrete values."""
 
@@ -67,5 +73,5 @@ class Condition(JsonModel):
     name: Optional[str] = None
     """Name of the condition."""
 
-    value: Optional[Union[StringConditionValue, NumericConditionValue]] = None
+    value: Optional[Union[NumericConditionValue, StringConditionValue]] = None
     """Value of the condition."""
--- a/nisystemlink/clients/spec/models/_condition.py
+++ b/nisystemlink/clients/spec/models/_condition.py
@@ -2,6 +2,7 @@ from enum import Enum
 from typing import List, Optional, Union
 
 from nisystemlink.clients.core._uplink._json_model import JsonModel
+from pydantic import StrictStr
 
 
 class ConditionType(Enum):
@@ -56,7 +57,7 @@ class StringConditionValue(ConditionValueBase):
     String conditions may only contain discrete lists of values.
     """
 
-    discrete: Optional[List[str]] = None
+    discrete: Optional[List[StrictStr]] = None
     """List of condition discrete values."""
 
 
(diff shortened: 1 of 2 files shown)


tests/integration/spec/test_spec.py
```diff
@@ -88,10 +89,9 @@ def create_specs_for_query(create_specs, product):
                 ),
                 Condition(
                     name="Supply Voltage",
-                    value=NumericConditionValue(
```
SSadaiyappan-NI, line 91: add discrete value to the previous numeric condition.

SSadaiyappan-NI (approved):

rbell517 (approved):

SSSantosh18 (approved):

SSadaiyappan-NI (approved):

tests/integration/spec/test_spec.py
```diff
@@ -298,6 +299,35 @@ def test__query_spec_projection_columns__columns_returned(
...
+            if response.specs[1].conditions
+            else None
+        )
+        condition_2 = (
+            response.specs[1].conditions[1].value
+            if response.specs[1].conditions
+            else None
+        )
+        assert isinstance(condition_1, NumericConditionValue)
+        assert isinstance(condition_2, StringConditionValue)
```
rbell517, line 329: Please also verify the `discrete` field's type as well in the numeric case and that we did not drop float precision.

nisystemlink/clients/spec/models/_condition.py
```diff
@@ -43,7 +44,10 @@ class NumericConditionValue(ConditionValueBase):
     range: Optional[List[ConditionRange]] = None
     """List of condition range values."""
 
-    discrete: Optional[List[float]] = None
+    # StrictFloat and StrictInt are used here because discrete is a common property for
+    # NumericConditionValue and StringConditionValue.
+    # And pydantic converts string of number into float/int by default if just float/int is used
+    discrete: Optional[List[Union[StrictFloat, StrictInt]]] = None
```
rbell517, line 50: Just to check my understanding here, StrictFloat will not accept integers but StrictInt will accept floats, according to the docs, so this ordering means it will prefer floats for floats without losing precision. If the other order was picked we'd always end up with integers and it would chop or round. Do I have that right? If so I think expanding your above comment to explain this would help.

rbell517 (approved):

nisystemlink/clients/spec/models/_condition.py
RSam-NI: `StrictInt` also won't accept anything else rather than integer values. In the pydantic validation, it is not accepting float values. Found it while testing manually. So, the order really doesn't matter here.

tests/integration/spec/test_spec.py
RSam-NI: Added assertions for verifying `discrete` field's type.

nisystemlink/clients/spec/models/_condition.py
SSSantosh18: image

StrictInt only allows int and subclasses of int. So, float will not be accepted by strictint as float is not a subclass of int

nisystemlink/clients/spec/models/_condition.py
```diff
@@ -43,7 +44,10 @@ class NumericConditionValue(ConditionValueBase):
     range: Optional[List[ConditionRange]] = None
     """List of condition range values."""
 
-    discrete: Optional[List[float]] = None
+    # StrictFloat and StrictInt are used here because discrete is a common property for
+    # NumericConditionValue and StringConditionValue.
+    # And pydantic converts string of number into float/int by default if just float/int is used
```
SSSantosh18, line 49:
```suggestion
    # If float/int is used, pydantic converts string of number into float/int by default when deserializing a StringCondition JSON and misinterprets it as NumericConditionValue type
```

nisystemlink/clients/spec/models/_condition.py
```diff
@@ -56,7 +60,10 @@ class StringConditionValue(ConditionValueBase):
     String conditions may only contain discrete lists of values.
     """
 
-    discrete: Optional[List[str]] = None
+    # StrictStr is used here because discrete is a common property for
+    # NumericConditionValue and StringConditionValue.
+    # And pydantic converts any datatype into string by default if just str is used.
```
SSSantosh18, line 65:
```suggestion
    # If str is used, pydantic converts any datatype into string by default when deserializing a NumericCondition JSON and misinterprets it as StringConditionValue type
```

nisystemlink/clients/spec/models/_condition.py
```diff
@@ -56,7 +60,10 @@ class StringConditionValue(ConditionValueBase):
     String conditions may only contain discrete lists of values.
     """
 
-    discrete: Optional[List[str]] = None
+    # StrictStr is used here because discrete is a common property for
```
SSSantosh18, line 63: Add the below comment above the `value` attribute in `Condition` class

`# Ideal approach is to set the dtype here as ConditionValue and use pydantic discriminator to deserialize/serialize the JSON into correct ConditionValue sub types. But Union of dtypes is used here as the discriminator field could be none when projection is used in query API`

SSSantosh18 (approved):

PR #86, opened 2024-12-04
feat: Add Client for Asset Management (Asset) API

sam-rishi:
- [x] This contribution adheres to [CONTRIBUTING.md](https://github.com/ni/nisystemlink-clients-python/blob/master/CONTRIBUTING.md).

### What does this Pull Request accomplish?

This PR adds API Client Module for `Asset` microservice in the `Asset Management` service.

### Why should this Pull Request be merged?

The `AssetManagementClient` makes it easier for users to interact with the endpoints of `Asset Management` service by just creating an instance of it.

### What testing has been done?

Automated integration tests are included.

API Link: [Swagger Link](https://dev-api.lifecyclesolutions.ni.com/niapis/?urls.primaryName=Asset+Management#/)

PR #134, opened 2025-05-15
feat: Add Client for Asset Management (Asset) create and query API 

aravindhan-ni:

What does this Pull Request accomplish?
This PR adds create and query API Client Module for Asset microservice in the Asset Management service.

Why should this Pull Request be merged?
The AssetManagementClient makes it easier for users to interact with the endpoints of create and query assets from Asset Management service by just creating an instance of it.

What testing has been done?

API Link: [Swagger Link](https://dev-api.lifecyclesolutions.ni.com/niapis/?urls.primaryName=Asset+Management#/)

### DIFF
--- a/examples/assetmanagement/assets.py
+++ b/examples/assetmanagement/assets.py
@@ -1,0 +1,20 @@
+from nisystemlink.clients.assetmanagement import AssetManagementClient
+from nisystemlink.clients.assetmanagement.models import QueryAssetRequest
+from nisystemlink.clients.core._http_configuration import HttpConfiguration
+
+server_configuration = HttpConfiguration(
+    server_uri="https://yourserver.yourcompany.com",
+    api_key="YourAPIKeyGeneratedFromSystemLink",
+)
+client = AssetManagementClient(configuration=server_configuration)
+
+# Query assets using id.
+queryRequest = QueryAssetRequest(
+    ids=["asset_ids"],
+    skip=0,
+    take=1,
+    descending=False,
+    calibratable_only=False,
+    returnCount=False,
+)
+queryResponse = client.query_assets(queryRequest)
--- a/nisystemlink/clients/assetmanagement/_asset_management_client.py
+++ b/nisystemlink/clients/assetmanagement/_asset_management_client.py
@@ -1,0 +1,64 @@
+"""Implementation of AssetManagementClient."""
+
+from typing import List, Optional
+
+from nisystemlink.clients import core
+from nisystemlink.clients.core._http_configuration import HttpConfiguration
+from nisystemlink.clients.core._uplink._base_client import BaseClient
+from nisystemlink.clients.core._uplink._methods import post
+from uplink import Field
+
+from . import models
+
+
+class AssetManagementClient(BaseClient):
+    def __init__(self, configuration: Optional[HttpConfiguration] = None):
+        """Initialize an instance.
+
+        Args:
+            configuration: Defines the web server to connect to and information about
+                how to connect. If not provided, the
+                :class:`HttpConfigurationManager <nisystemlink.clients.core.HttpConfigurationManager>`
+                is used to obtain the configuration.
+
+        Raises:
+            ApiException: if unable to communicate with the AssetManagement Service.
+        """
+        if configuration is None:
+            configuration = core.HttpConfigurationManager.get_configuration()
+
+        super().__init__(configuration, base_path="/niapm/v1/")
+
+    @post("assets", args=[Field("assets")])
+    def create_assets(
+        self, assets: List[models.AssetCreateRequest]
+    ) -> models.AssetsCreatePartialSuccessResponse:
+        """Create Assets.
+
+        Args:
+            assets: an array of assets that should be created.
+
+        Returns:
+            Response object with array of created assets and array of failed assets.
+
+        Raises:
+            ApiException: if unable to communicate with the `nispec` service or if there are invalid
+            arguments.
+        """
+        ...
+
+    @post("query-assets")
+    def query_assets(self, query: models.QueryAssetRequest) -> models.AssetsResponse:
+        """Query Assets.
+
+        Args:
+            query: an object containing filters to apply when retrieving assets.
+
+        Returns:
+            Assets Response containing the assets satisfying the query and the total count of matching assets.
+
+        Raises:
+            ApiException: if unable to communicate with the `nispec` service or if there are invalid
+            arguments.
+        """
+        ...
--- a/nisystemlink/clients/assetmanagement/models/_asset.py
+++ b/nisystemlink/clients/assetmanagement/models/_asset.py
@@ -1,0 +1,301 @@
+from datetime import datetime
+from enum import Enum
+from typing import Dict, List, Optional
+
+from nisystemlink.clients.core._uplink._json_model import JsonModel
+
+
+class AssetBusType(Enum):
+    """All supported bus types for an asset."""
+
+    BUILT_IN_SYSTEM = "BUILT_IN_SYSTEM"
+
+    PCI_PXI = "PCI_PXI"
+
+    USB = "USB"
+
+    GPIB = "GPIB"
+
+    VXI = "VXI"
+
+    SERIAL = "SERIAL"
+
+    TCP_IP = "TCP_IP"
+
+    CRIO = "CRIO"
+
+    SCXI = "SCXI"
+
+    CDAQ = "CDAQ"
+
+    SWITCH_BLOCK = "SWITCH_BLOCK"
+
+    SCC = "SCC"
+
+    FIRE_WIRE = "FIRE_WIRE"
+
+    ACCESSORY = "ACCESSORY"
+
+    CAN = "CAN"
+
+    SWITCH_BLOCK_DEVICE = "SWITCH_BLOCK_DEVICE"
+
+    SLSC = "SLSC"
+
+
+class AssetType(Enum):
+    """All supported asset types."""
+
+    GENERIC = "GENERIC"
+
+    DEVICE_UNDER_TEST = "DEVICE_UNDER_TEST"
+
+    FIXTURE = "FIXTURE"
+
+    SYSTEM = "SYSTEM"
+
+
+class AssetDiscoveryType(Enum):
+    """All discovery types."""
+
+    MANUAL = "MANUAL"
+
+    AUTOMATIC = "AUTOMATIC"
+
+
+class TemperatureSensor(JsonModel):
+    """Temperature sensor information."""
+
+    name: Optional[str] = None
+    """Gets or sets sensor name."""
+
+    reading: float
+    """Gets or sets sensor reading."""
+
+
+class AssetPresence(Enum):
+    """Status of an asset's presence in a system."""
+
+    INITIALIZING = "INITIALIZING"
+
+    UNKNOWN = "UNKNOWN"
+
+    NOT_PRESENT = "NOT_PRESENT"
+
+    PRESENT = "PRESENT"
+
+
+class SystemConnection(Enum):
+    """Whether or not the minion is connected to the server and has updated the server with its data.
+    To maintain compatibility with previous versions of SystemLink, the values
+    [APPROVED, UNSUPPORTED, ACTIVATED] are considered equivalent to DISCONNECTED and
+    [CONNECTED_UPDATE_PENDING, CONNECTED_UPDATE_SUCCESSFUL, CONNECTED_UPDATE_FAILED] are equivalent to CONNECTED.
+    """
+
+    APPROVED = "APPROVED"
+
+    DISCONNECTED = "DISCONNECTED"
+
+    CONNECTED_UPDATE_PENDING = "CONNECTED_UPDATE_PENDING"
+
+    CONNECTED = "CONNECTED"
+
+    CONNECTED_UPDATE_FAILED = "CONNECTED_UPDATE_FAILED"
+
+    UNSUPPORTED = "UNSUPPORTED"
+
+    ACTIVATED = "ACTIVATED"
+
+    CONNECTED_UPDATE_SUCCESSFUL = "CONNECTED_UPDATE_SUCCESSFUL"
+
+
+class AssetPresenceWithSystemConnection(JsonModel):
+    """Model for the presence of an asset and the connection of the system in which it resides."""
+
+    asset_presence: AssetPresence
+    """Gets or sets the status of an asset's presence in a system."""
+
+    system_connection: Optional[SystemConnection] = None
+    """Gets or sets whether or not the minion is connected to the server and has updated the server with its data."""
+
+
+class AssetLocation(JsonModel):
+    """Model for information about the asset location, presence and the connection status of the system."""
+
+    minion_id: Optional[str] = None
+    """Gets or sets identifier of the minion where the asset is located."""
+
+    physical_location: Optional[str] = None
+    """Gets or sets the physical location of the asset."""
+
+    parent: Optional[str] = None
+    """Gets or sets the parent of the asset."""
+
+    resource_uri: Optional[str] = None
+    """Gets or sets identifier of a resource."""
+
+    slot_number: Optional[int] = None
+    """Gets or sets the number of the slot in which the asset is located."""
+
+    state: AssetPresenceWithSystemConnection
+    """Presence of an asset and the connection of the system in which it resides."""
+
+
+class SelfCalibration(JsonModel):
+    temperature_sensors: Optional[List[TemperatureSensor]] = None
+    """Gets or sets an array of temperature sensor information."""
+
+    is_limited: Optional[bool] = None
+    """Gets or sets whether the last self-calibration of the asset was a limited calibration."""
+
+    date: datetime
+    """Gets or sets ISO-8601 formatted timestamp specifying the last date the asset was self-calibrated."""
+
+
+class CalibrationStatus(Enum):
+    """Calibration category the asset belongs to based on the next due calibration date."""
+
+    OK = "OK"
+
+    APPROACHING_RECOMMENDED_DUE_DATE = "APPROACHING_RECOMMENDED_DUE_DATE"
+
+    PAST_RECOMMENDED_DUE_DATE = "PAST_RECOMMENDED_DUE_DATE"
+
+    OUT_FOR_CALIBRATION = "OUT_FOR_CALIBRATION"
+
+
+class CalibrationMode(Enum):
+    """Whether SystemLink automatically discovered the calibration data for an asset or if it was manually entered."""
+
+    AUTOMATIC = "AUTOMATIC"
+
+    MANUAL = "MANUAL"
+
+
+class ExternalCalibration(JsonModel):
+    temperature_sensors: Optional[List[TemperatureSensor]] = None
+    """Gets or sets an array of temperature sensor information."""
+
+    is_limited: Optional[bool] = None
+    """Gets or sets whether the last external calibration of the asset was a limited calibration."""
+
+    date: str
+    """Gets or sets ISO-8601 formatted timestamp specifying the last date the asset was externally calibrated."""
+
+    recommended_interval: int
+    """Gets or sets the manufacturer's recommended calibration interval in months."""
+
+    next_recommended_date: str
+    """Gets or sets ISO-8601 formatted timestamp specifying the recommended date for the next external calibration."""
+
+    next_custom_due_date: Optional[str] = None
+    """Gets or sets ISO-8601 formatted timestamp specifying the date for the next external calibration."""
+
+    resolved_due_date: Optional[str] = None
+    """Gets ISO-8601 formatted timestamp specifying the resolved due date for external calibration."""
+
+    comments: Optional[str] = None
+    """Gets or sets calibration comments provided by an operator."""
+
+    entry_type: Optional[CalibrationMode] = None
+    """Gets or sets whether automatically discovered the calibration data for an asset or manually entered."""
+
+
+class Asset(JsonModel):
+    """Model for an object describing an asset with all of its properties."""
+
+    model_name: Optional[str] = None
+    """Gets or sets model name of the asset."""
+
+    model_number: int
+    """Gets or sets model number of the asset."""
+
+    serial_number: Optional[str] = None
+    """Gets or sets serial number of the asset."""
+
+    vendor_name: Optional[str] = None
+    """Gets or sets vendor name of the asset."""
+
+    vendor_number: int
+    """Gets or sets vendor number of the asset."""
+
+    bus_type: AssetBusType
+    """Gets or sets all supported bus types for an asset."""
+
+    name: Optional[str] = None
+    """Gets or sets name of the asset."""
+
+    asset_type: AssetType
+    """Gets or sets all supported asset types."""
+
+    discovery_type: AssetDiscoveryType
+    """Gets or sets the discovery type."""
+
+    firmware_version: Optional[str] = None
+    """Gets or sets firmware version of the asset."""
+
+    hardware_version: Optional[str] = None
+    """Gets or sets hardware version of the asset."""
+
+    visa_resource_name: Optional[str] = None
+    """Gets or sets VISA resource name of the asset."""
+
+    temperature_sensors: Optional[List[TemperatureSensor]] = None
+    """Gets or sets an array of temperature sensor information."""
+
+    supports_self_calibration: bool
+    """Gets or sets whether the asset supports self-calibration."""
+
+    supports_external_calibration: bool
+    """Gets or sets whether the asset supports external calibration."""
+
+    custom_calibration_interval: Optional[int] = None
+    """Gets or sets the interval represented in months used for computing calibration due date."""
+
+    self_calibration: Optional[SelfCalibration] = None
+    """Gets or sets the last self-calibration of the asset."""
+
+    is_ni_asset: bool
+    """Gets or sets whether this asset is an NI asset (true) or a third-party asset (false)."""
+
+    id: Optional[str] = None
+    """Gets or sets unique identifier of the asset."""
+
+    location: AssetLocation
+    """Model for information about the asset location, presence and the connection status of the system"""
+
+    calibration_status: Optional[CalibrationStatus] = None
+    """Gets or sets the calibration category the asset belongs to based on the next due calibration date."""
+
+    is_system_controller: bool
+    """Gets or sets whether this asset represents a System Controller."""
+
+    external_calibration: Optional[ExternalCalibration] = None
+    """Gets or sets the last external calibration of the asset."""
+
+    workspace: Optional[str] = None
+    """Gets or sets the ID of the workspace."""
+
+    properties: Dict[str, str]
+    """	Gets or sets key-value-pair metadata associated with an asset."""
+
+    keywords: Optional[List[str]] = None
+    """Gets or sets words or phrases associated with an asset."""
+
+    last_updated_timestamp: str
+    """Gets or sets ISO-8601 formatted timestamp specifying the last date that the asset has had a property update."""
+
+    file_ids: Optional[List[str]] = None
+    """Gets or sets all files linked to the asset."""
+
+    supports_self_test: bool
+    """Gets or sets whether the asset supports self-test."""
+
+    supports_reset: bool
+    """Gets or sets whether the asset supports reset."""
+
+    part_number: Optional[str] = None
+    """Gets or sets part number of the asset."""
+
+    out_for_calibration: Optional[bool] = None
+    """Get or set whether the asset is out for calibration."""
--- a/nisystemlink/clients/assetmanagement/models/_asset_create.py
+++ b/nisystemlink/clients/assetmanagement/models/_asset_create.py
@@ -1,0 +1,113 @@
+from typing import Dict, List, Optional
+
+from nisystemlink.clients.core._api_error import ApiError
+from nisystemlink.clients.core._uplink._json_model import JsonModel
+
+from ._asset import (
+    Asset,
+    AssetBusType,
+    AssetDiscoveryType,
+    AssetLocation,
+    AssetType,
+    ExternalCalibration,
+    SelfCalibration,
+    TemperatureSensor,
+)
+
+
+class AssetCreateRequest(JsonModel):
+    """Model for an object describing the properties for creating an asset."""
+
+    model_name: Optional[str] = None
+    """Gets or sets model name of the asset."""
+
+    model_number: Optional[int] = None
+    """Gets or sets model number of the asset."""
+
+    serial_number: Optional[str] = None
+    """Gets or sets serial number of the asset."""
+
+    vendor_name: Optional[str] = None
+    """Gets or sets vendor name of the asset."""
+
+    vendor_number: Optional[int] = None
+    """Gets or sets vendor number of the asset."""
+
+    bus_type: AssetBusType
+    """Gets or sets all supported bus types for an asset."""
+
+    name: Optional[str] = None
+    """Gets or sets name of the asset."""
+
+    asset_type: Optional[AssetType] = None
+    """Gets or sets all supported asset types."""
+
+    firmware_version: Optional[str] = None
+    """Gets or sets firmware version of the asset."""
+
+    hardware_version: Optional[str] = None
+    """Gets or sets hardware version of the asset."""
+
+    visa_resource_name: Optional[str] = None
+    """Gets or sets VISA resource name of the asset."""
+
+    temperature_sensors: Optional[List[TemperatureSensor]] = None
+    """Gets or sets an array of temperature sensor information."""
+
+    supports_self_calibration: Optional[bool] = None
+    """Gets or sets whether the asset supports self-calibration."""
+
+    supports_external_calibration: Optional[bool] = None
+    """Gets or sets whether the asset supports external calibration."""
+
+    custom_calibration_interval: Optional[int] = None
+    """Gets or sets the interval represented in months used for computing calibration due date."""
+
+    self_calibration: Optional[SelfCalibration] = None
+    """Gets or sets the last self-calibration of the asset."""
+
+    is_ni_asset: Optional[bool] = None
+    """Gets or sets whether this asset is an NI asset (true) or a third-party asset (false)."""
+
+    workspace: Optional[str] = None
+    """Gets or sets the ID of the workspace."""
+
+    location: AssetLocation
+    """Model for information about the asset location, presence and the connection status of the system"""
+
+    external_calibration: Optional[ExternalCalibration] = None
+    """Gets or sets the last external calibration of the asset."""
+
+    properties: Optional[Dict[str, str]] = None
+    """	Gets or sets key-value-pair metadata associated with an asset."""
+
+    keywords: Optional[List[str]] = None
+    """Gets or sets words or phrases associated with an asset."""
+
+    discovery_type: Optional[AssetDiscoveryType] = None
+    """Gets or sets the discovery type."""
+
+    file_ids: Optional[List[str]] = None
+    """Gets or sets all files linked to the asset."""
+
+    supports_self_test: Optional[bool] = None
+    """Gets or sets whether the asset supports self-test."""
+
+    supports_reset: Optional[bool] = None
+    """Gets or sets whether the asset supports reset."""
+
+    part_number: Optional[str] = None
+    """Gets or sets part number of the asset."""
+
+
+class AssetsCreatePartialSuccessResponse(JsonModel):
+    """Model for create Assets Partial Success Response."""
+
+    error: Optional[ApiError] = None
+    """Gets or sets error information if the request failed."""
+
+    assets: Optional[List[Asset]] = None
+    """Gets or sets array of created assets."""
+
+    failed: Optional[List[AssetCreateRequest]] = None
+    """Gets or sets array of assets create requests that failed."""
--- a/nisystemlink/clients/assetmanagement/models/_query_assets.py
+++ b/nisystemlink/clients/assetmanagement/models/_query_assets.py
@@ -1,0 +1,47 @@
+from enum import Enum
+from typing import List, Optional
+
+from nisystemlink.clients.core._uplink._json_model import JsonModel
+
+
+class OrderBy(Enum):
+    """Field by which assets can be ordered/sorted. If OrderBy is not specified, no sorting will applied."""
+
+    LAST_UPDATED_TIMESTAMP = "LAST_UPDATED_TIMESTAMP"
+
+
+class QueryAssetRequest(JsonModel):
+    """Model for object containing filters to apply when retrieving assets."""
+
+    ids: Optional[List[str]] = None
+    """Gets or sets identifiers of the assets to be retrieved."""
+
+    projection: Optional[str] = None
+    """
+    Gets or sets the projection to be used when retrieving the assets. If not specified,
+    all properties will be returned.
+    """
+
+    skip: int
+    """Gets or sets the number of resources to skip in the result when paging."""
+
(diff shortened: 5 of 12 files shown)


tests/integration/assetmanagement/test_asset_management.py
```diff
@@ -0,0 +1,105 @@
...
+@pytest.mark.integration
+@pytest.mark.enterprise
+class TestAssetManagement:
+
+    def test__query_assets_with_take_value__returns_specific_number_of_assets(
+        self, client: AssetManagementClient, asset_create: List[AssetCreateRequest]
+    ):
+        asset_create[0].model_number = 2005
+        create_assets_response: AssetsCreatePartialSuccessResponse = (
+            client.create_assets(assets=asset_create)
```
rbell517, line 92: I will have time to dig into this review more tomorrow but to give you quick feedback on probably the more impactful part of this re:my comment in the user story -- 

If you are making backend entities in the integration tests you must include a step to clean those entities up during test cleanup. See the other client test cases for examples of this. That will mean you need to include a delete API in your client if you want to test like this. The assets delete API is dead simple, so I think its probably simpler to go ahead and add it instead of paring back the create support and reworking your tests.

But if you'd really rather only support query, we have [examples](https://github.com/ni/nisystemlink-clients-python/blob/d987b4071c8dee82d36247553dccfdc883fad0a1/tests/integration/notebook/test_notebook_client.py#L508-L560)in the notebook client for how you can mock the API response using `@requests.activate`. Since this isn't testing the actual JSON returned from the API it is not perfect and bugs have snuck in on these tests, so I would push for only doing this if necessary, like notebook exec where actually creating executions is costly and slow.

nisystemlink/clients/assetmanagement/_asset_management_client.py
```diff
@@ -0,0 +1,80 @@
...
+        super().__init__(configuration, base_path="/niapm/v1/")
+
+    @post("assets", args=[Field("assets")])
+    def create_assets(
+        self, assets: List[models.AssetCreateRequest]
+    ) -> models.AssetsCreatePartialSuccessResponse:
+        """Create Assets.
+
+        Args:
+            assets: an array of assets that should be created.
```
MuthuRajan-S, line 39:
```suggestion
            assets: Array of assets that should be created.
```
nit

nisystemlink/clients/assetmanagement/_asset_management_client.py
```diff
@@ -0,0 +1,80 @@
...
+
+        Returns:
+            Response object with array of created assets and array of failed assets.
```
MuthuRajan-S, line 42:
```suggestion
            AssetsCreatePartialSuccessResponse: Array of created assets and array of failed assets.
```

MuthuRajan-S: Would be good to return with its type, refer to [artifact client](https://github.com/ni/nisystemlink-clients-python/blob/master/nisystemlink/clients/artifact/_artifact_client.py).

Apply elsewhere.

examples/assetmanagement/assets.py
```diff
@@ -0,0 +1,92 @@
...
+            supports_reset=True,
+            partNumber="A1234 B5",
+        )
+    ]
+
+    response = client.create_assets(assets=assets)
+    return response
+
+
+server_configuration = HttpConfiguration(
```
MuthuRajan-S, line 71: Let's move server_configuration  and client instantiating to top of this file after import, basically go by the order, helps users in understanding.

examples/assetmanagement/assets.py
```diff
@@ -0,0 +1,92 @@
...
+    server_uri="https://yourserver.yourcompany.com",
+    api_key="YourAPIKeyGeneratedFromSystemLink",
+)
+client = AssetManagementClient(configuration=server_configuration)
+
+# Create an asset.
+createAssetResponse = create_an_asset()
```
MuthuRajan-S, line 78: Here, the query looks clean, defining request body and calling api in first level of the example file, but create does everything in same private function. Let's format in this way,

```suggestion
create_assets_request = [ ... ]
createAssetResponse = client.create_assets(assets=create_assets_request)
```

examples/assetmanagement/assets.py
```diff
@@ -0,0 +1,92 @@
...
+
+# Query assets using id.
+queryRequest = QueryAssetRequest(
+    ids=[createAssetResponse.assets[0].id],
```
MuthuRajan-S, line 82: DRY : Let's use a local variable named created_asset_id and store for first Id and use it in the query request + Use it for delete.

nisystemlink/clients/assetmanagement/models/_query_assets.py
```diff
@@ -0,0 +1,47 @@
...
+    order_by: Optional[OrderBy] = None
+    """Field by which assets can be ordered/sorted. If OrderBy is not specified, no sorting will applied."""
+
+    descending: Optional[bool] = None
+    """Whether to return the assets in the descending order. If OrderBy is not specified, this property is ignored."""
+
+    calibratable_only: bool
+    """Gets or sets whether to generate a report with calibrated asset specific columns:"""
+
+    returnCount: bool
```
MuthuRajan-S, line 40: I guess, all the fields are optional and have a default value handled by the service itself, if so, we can update accordingly.

nisystemlink/clients/assetmanagement/models/_asset_create.py
```diff
@@ -0,0 +1,113 @@
...
+    AssetDiscoveryType,
+    AssetLocation,
+    AssetType,
+    ExternalCalibration,
+    SelfCalibration,
+    TemperatureSensor,
+)
+
+
+class AssetCreateRequest(JsonModel):
```
MuthuRajan-S, line 18: The file names seems odd,

image

Let's be consistent with this format <action>_<entity>_<request/response>. Refer to [TMS client models](https://github.com/ni/nisystemlink-clients-python/tree/master/nisystemlink/clients/testmonitor/models).

Also, let's create dedicated DTOs / models for request and response.

nisystemlink/clients/assetmanagement/models/_asset.py
```diff
@@ -0,0 +1,301 @@
+from datetime import datetime
```
MuthuRajan-S, line 1: This class is bulky and handles more responsibilities, let's split it into multiple files and have dedicated responsibilities for them.

To me, it mostly appear as a Constants class, where you have set of constants. Better have a folder as Constants and remove it from Models (which is to define API models).

nisystemlink/clients/assetmanagement/models/_asset.py
```diff
@@ -0,0 +1,301 @@
...
+
+    SWITCH_BLOCK_DEVICE = "SWITCH_BLOCK_DEVICE"
+
+    SLSC = "SLSC"
+
+
+class AssetType(Enum):
+    """All supported asset types."""
+
+    GENERIC = "GENERIC"
```
MuthuRajan-S, line 49: Use [enum.auto()](https://www.geeksforgeeks.org/enum-auto-in-python/), apply other places as well.

tests/integration/assetmanagement/test_asset_management.py
```diff
@@ -0,0 +1,159 @@
...
+    ExternalCalibration,
+    QueryAssetRequest,
+    SelfCalibration,
+    TemperatureSensor,
+)
+from nisystemlink.clients.core._http_configuration import HttpConfiguration
+
+
+@pytest.fixture(scope="class")
+def asset_create() -> List[AssetCreateRequest]:
```
MuthuRajan-S, line 24: Naming can be consistent, refer to the other comment.

MuthuRajan-S (changes requested):
Overall, implementation looks good, needs a bit of clean up in classes + renames.

nisystemlink/clients/assetmanagement/_asset_management_client.py
aravindhan-ni: Updated as suggested.

nisystemlink/clients/assetmanagement/_asset_management_client.py
aravindhan-ni: Updated as suggested.

examples/assetmanagement/assets.py
aravindhan-ni: Updated as suggested.

examples/assetmanagement/assets.py
aravindhan-ni: Updated as suggested.

examples/assetmanagement/assets.py
aravindhan-ni: Updated as suggested.

nisystemlink/clients/assetmanagement/models/_asset.py
aravindhan-ni: Updated as suggested.

aravindhan-ni: Revert back - getting validation errors

nisystemlink/clients/assetmanagement/models/_asset_create.py
aravindhan-ni: Updated file names

nisystemlink/clients/assetmanagement/models/_asset.py
```diff
@@ -0,0 +1,112 @@
...
+from ._asset_types import AssetBusType, AssetDiscoveryType, AssetType
+
+
+class Asset(JsonModel):
+    """Model for an object describing an asset with all of its properties."""
+
+    model_name: Optional[str] = None
+    """Gets or sets model name of the asset."""
+
+    model_number: int
```
rbell517, line 21: We are supporting projections, so all the query fields need to be optional. This needs to be applied down the model as well.
```suggestion
    model_number: Optional[int] = None
```
We don't need to apply this within the sub-models of the query response, like for location details, just at the top-level fields (what we will explicitly support in the projection enum, see my other comment).

nisystemlink/clients/assetmanagement/models/_asset_calibration.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+class ExternalCalibration(JsonModel):
+    temperature_sensors: Optional[List[TemperatureSensor]] = None
+    """Gets or sets an array of temperature sensor information."""
+
+    is_limited: Optional[bool] = None
+    """Gets or sets whether the last external calibration of the asset was a limited calibration."""
+
+    date: str
```
rbell517, line 54: All datetime fields need to use the proper python datetime type. Pydantic will convert this for you automatically, so you'll just need to adjust the type.

rbell517:
```suggestion
    date: datetime
```

nisystemlink/clients/assetmanagement/models/_query_assets_request.py
```diff
@@ -0,0 +1,47 @@
...
+    LAST_UPDATED_TIMESTAMP = "LAST_UPDATED_TIMESTAMP"
+
+
+class QueryAssetsRequest(JsonModel):
+    """Model for object containing filters to apply when retrieving assets."""
+
+    ids: Optional[List[str]] = None
+    """Gets or sets identifiers of the assets to be retrieved."""
+
+    projection: Optional[str] = None
```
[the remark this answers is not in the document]
rbell517: We are going to need to decide which fields to allow projection for. The API does appear to allow projections down within the sub fields of the asset model, e.g. `new(name, location.state.assetPresence)` to get the name and only the `assetPresence` field in the location part of the model, not the whole location. I personally don't feel like we need to go that far and can just support the fields directly under the asset model and not these sub fields. That will make the projection enum simpler. 

@CiprianAnton I'm curious your opinion on this point.

tests/integration/assetmanagement/test_asset_management.py
```diff
@@ -0,0 +1,166 @@
...
+            keywords=["Keyword1"],
+            discovery_type=AssetDiscoveryType.AUTOMATIC,
+            file_ids=["608a5684800e325b48837c2a"],
+            supports_self_test=True,
+            supports_reset=True,
+            part_number="A1234 B5",
+        )
+    ]
+
+    return assets
```
rbell517, line 72: The other fixtures like this actually create the resource within the fixture and then `yield` to allow the same fixture that created the resource to delete the resource as part of the test cleanup. That is how we should implement the cleanup now that we have delete rather than just adding deletes at the end of the test body. That will ensure in cases where the test fails the test cleanup will still try to delete anything it created. The specs integration tests are a good example of this.

tests/integration/assetmanagement/test_asset_management.py
```diff
@@ -0,0 +1,166 @@
...
+        assert asset_id is not None
+
+        delete_response = client.delete_assets(ids=[asset_id])
+
+        assert delete_response is not None
+        assert delete_response.ids is not None
+        assert len(delete_response.ids) == 1
+        assert delete_response.ids[0] == asset_id
+
+    def test__query_assets_with_take_value__returns_specific_number_of_assets(
```
rbell517, line 135: Make sure to also include tests that cover the projection support you add.

tests/integration/assetmanagement/test_asset_management.py
```diff
@@ -0,0 +1,166 @@
...
+        )
+        asset_id = (
+            create_response.assets[0].id
+            if create_response.assets and create_response.assets[0].id
+            else None
+        )
+
+        assert asset_id is not None
+
+        query_request = QueryAssetsRequest(
```
rbell517, line 152: Your query should be filtering down to the asset you created, not just getting any old asset back.

tests/integration/assetmanagement/test_asset_management.py
```diff
@@ -0,0 +1,166 @@
...
+def create_assets_request() -> List[AssetCreateRequest]:
+    """Fixture to create an AssetCreateRequest object."""
+    assets = [
+        AssetCreateRequest(
+            model_name="NI PXIe-6368",
+            serial_number="01BB877A",
+            vendor_name="NI",
+            vendor_number="4244",
+            bus_type=AssetBusType.ACCESSORY,
+            name="PCISlot2",
```
rbell517, line 33: For these strings we can pick we should pick values that make it very obvious that this asset is from the python integration tests, so we know if they get leaked that they are safe to remove. E.g. name="Python Integration Tests DUT"

nisystemlink/clients/assetmanagement/models/_query_assets_response.py
```diff
@@ -0,0 +1,19 @@
...
+from nisystemlink.clients.core._api_error import ApiError
+from nisystemlink.clients.core._uplink._json_model import JsonModel
+
+from ._asset import Asset
+
+
+class QueryAssetsResponse(JsonModel):
+    """Model for assets Response containing the assets satisfying the query and the total count of matching assets."""
+
+    error: Optional[ApiError] = None
```
rbell517, line 12: Best I can tell the query response will never have an API error field, despite the swagger. It will only have an error when the status code is an error status code, which is separately handled by the base client and wouldn't need to be in this model.

nisystemlink/clients/assetmanagement/_asset_management_client.py
```diff
@@ -0,0 +1,83 @@
...
+from nisystemlink.clients import core
+from nisystemlink.clients.core._http_configuration import HttpConfiguration
+from nisystemlink.clients.core._uplink._base_client import BaseClient
+from nisystemlink.clients.core._uplink._methods import post
+from uplink import Field
+
+from . import models
+
+
+class AssetManagementClient(BaseClient):
```
rbell517, line 14: You want to include the retry logic other clients have 
```suggestion
@retry(
    when=retry.when.status(408, 429, 502, 503, 504),
    stop=retry.stop.after_attempt(5),
    on_exception=retry.CONNECTION_ERROR,
)
class AssetManagementClient(BaseClient):
```

nisystemlink/clients/assetmanagement/models/_asset.py
```diff
@@ -0,0 +1,112 @@
...
+    supports_external_calibration: bool
+    """Gets or sets whether the asset supports external calibration."""
+
+    custom_calibration_interval: Optional[int] = None
+    """Gets or sets the interval represented in months used for computing calibration due date."""
+
+    self_calibration: Optional[SelfCalibration] = None
+    """Gets or sets the last self-calibration of the asset."""
+
+    is_n_i_asset: bool
```
rbell517, line 69:
```suggestion
    is_NI_asset: bool
```

You might need to force a specific alias to get Pydantic to parse this properly (plus projection support)

```suggestion
    is_NI_asset: Optional[bool] = Field(alias="isNIAsset", default=None)
```

nisystemlink/clients/assetmanagement/models/_asset_create_request.py
```diff
@@ -0,0 +1,98 @@
...
+    serial_number: Optional[str] = None
+    """Gets or sets serial number of the asset."""
+
+    vendor_name: Optional[str] = None
+    """Gets or sets vendor name of the asset."""
+
+    vendor_number: Optional[int] = None
+    """Gets or sets vendor number of the asset."""
+
+    bus_type: AssetBusType
```
rbell517, line 34: Bus type should also be optional

nisystemlink/clients/assetmanagement/models/_asset_create_request.py
```diff
@@ -0,0 +1,98 @@
...
+    ExternalCalibration,
+    SelfCalibration,
+    TemperatureSensor,
+)
+
+
+class AssetCreateRequest(JsonModel):
+    """Model for an object describing the properties for creating an asset."""
+
+    model_name: Optional[str] = None
```
rbell517, line 19: There is a caveat to the model_name being optional. The service requires the asset to be uniquely identifiable by a combination of modelName, modelNumber, serialNumber, vendorName, vendorNumber, busType, and minionId. We do have the ability to write a Pydantic validator that would enforce this in the client. That validation seems like a nice to have, but I don't like the idea of duplicating service validation logic into the client and keeping those two in sync if the service behavior ever needs to change. So all that said, I don't think you need to change anything, just adding the context and opening up for comment. FYI @CiprianAnton

nisystemlink/clients/assetmanagement/models/_asset_calibration.py
rbell517: Searching for ISO-8601 in your files will help

rbell517 (changes requested):

nisystemlink/clients/assetmanagement/models/_asset_create.py
MuthuRajan-S: I guess, this is not fully addressed, there are partial create classes still renames renaming, refer to its actual model in C# class and add it here. (I can't link the source code, as this is a public repo.)

nisystemlink/clients/assetmanagement/models/_query_assets_request.py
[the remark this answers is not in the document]
rbell517: The goal here and with nbexec is that we wanted a more approachable, pythonic API so you can just use the enum values to know what projections are possible. There is also a consistency question with the other APIs that are enums at the API level and do not support the `new()` style. Nbexec this wasn't much of a problem because its nested object field is just `source` which is small. Assets' nested object fields are a lot larger, and so the value of projecting into these objects is higher, but its still unclear to me whether it needs to be first class in an enum.

The enum approach could still be used for sub-fields, you just need a larger number of enum values

```python
class AssetsProjection(enum, str):
    ...
    LOCATION = auto()
    LOCATION_MINION_ID = auto()
    LOCATION_PHYSICAL_LOCATION = auto()
    LOCATION_STATE = auto()
    LOCATION_STATE_ASSET_PRESENCE = auto()
```

The other option is to support the first level fields in the projection enum but give an option to escape to the string to get a more specific projection, which we could use a union to do

```python
class QueryAssetsRequest(JsonModel):
    ...
    projection: Optional[Union[List[AssetsProjection], string]] = None
```

If the projection is the enum list option, we convert that to the string. If it is a string, we pass it on as is.

We have a policy to do major version revisions if we include breaking changes, but I think changing the field to `Union[List[Projection], string]` to allow either depending on what the caller wants would not be considered an API break. That would mean we can add that later if we want, for both assets and nbexec (and systems shortly). We can start now with just the first-level fields in the enum and decide later to add more enum values for the sub-fields, or we can add the string input option, or both.

The net of this for me is starting with the first-level fields in a projection enum to be consistent and approachable, and then later as the need arises we consider expansion as described above.

examples/assetmanagement/assets.py
```diff
@@ -0,0 +1,91 @@
...
+        temperature_sensors=[TemperatureSensor(name="Sensor0", reading=25.8)],
+        supports_self_calibration=True,
+        supports_external_calibration=True,
+        custom_calibration_interval=24,
+        self_calibration=SelfCalibration(
+            temperature_sensors=[TemperatureSensor(name="Sensor0", reading=25.8)],
+            is_limited=False,
+            date="2022-06-07T18:58:05.000Z",
+        ),
+        is_n_i_asset=True,
```
CiprianAnton, line 46: is n_i instead of ni intentional?

nisystemlink/clients/assetmanagement/models/_asset_create.py
aravindhan-ni: Referred and updated

examples/assetmanagement/assets.py
```diff
@@ -0,0 +1,91 @@
...
+        workspace="846e294a-a007-47ac-9fc2-fac07eab240e",
+        location=AssetLocation(
+            state=AssetPresenceWithSystemConnection(
```
CiprianAnton, line 49: System Connection is not exposed in create/update assets request. Suggestion for the class name: `AssetPresenceModel`

examples/assetmanagement/assets.py
```diff
@@ -0,0 +1,91 @@
...
+            temperature_sensors=[TemperatureSensor(name="Sensor0", reading=25.8)],
+            date="2022-06-07T18:58:05.000Z",
+            recommended_interval=10,
+            next_recommended_date="2023-11-14T20:42:11.583Z",
+            next_custom_due_date="2024-11-14T20:42:11.583Z",
+            resolved_due_date="2022-06-07T18:58:05.000Z",
+        ),
+        properties={"Key1": "Value1"},
+        keywords=["Keyword1"],
+        discovery_type=AssetDiscoveryType.AUTOMATIC,
```
CiprianAnton, line 63: Let's use MANUAL for the example, as AUTOMATIC stands for those assets reported by drivers

tests/integration/assetmanagement/test_asset_management.py
aravindhan-ni: Updated

nisystemlink/clients/assetmanagement/models/_asset_calibration.py
aravindhan-ni: Updated

nisystemlink/clients/assetmanagement/models/_asset_location.py
```diff
@@ -0,0 +1,62 @@
...
+    """Whether or not the minion is connected to the server and has updated the server with its data.
+    To maintain compatibility with previous versions of SystemLink, the values
+    [APPROVED, UNSUPPORTED, ACTIVATED] are considered equivalent to DISCONNECTED and
+    [CONNECTED_UPDATE_PENDING, CONNECTED_UPDATE_SUCCESSFUL, CONNECTED_UPDATE_FAILED] are equivalent to CONNECTED.
+    """
+
+    APPROVED = "APPROVED"
+    DISCONNECTED = "DISCONNECTED"
+    CONNECTED_UPDATE_PENDING = "CONNECTED_UPDATE_PENDING"
+    CONNECTED = "CONNECTED"
```
CiprianAnton, line 26: Is there a way to deprecated fields? Connected and Disconnected are the supported fields, the rest is legacy

tests/integration/assetmanagement/test_asset_management.py
aravindhan-ni: Updated as suggested

nisystemlink/clients/assetmanagement/models/_query_assets_response.py
aravindhan-ni: Removed error field

nisystemlink/clients/assetmanagement/_asset_management_client.py
aravindhan-ni: Added

nisystemlink/clients/assetmanagement/models/_create_asset_request.py
```diff
@@ -0,0 +1,98 @@
...
+    self_calibration: Optional[SelfCalibration] = None
+    """Gets or sets the last self-calibration of the asset."""
+
+    is_NI_asset: Optional[bool] = None
+    """Gets or sets whether this asset is an NI asset (true) or a third-party asset (false)."""
+
+    workspace: Optional[str] = None
+    """Gets or sets the ID of the workspace."""
+
+    location: AssetLocation
```
CiprianAnton, line 73: You need to diferentiate location model from query vs update/create. The query also includes system connection, but the update/create not

nisystemlink/clients/assetmanagement/models/_create_assets_partial_success_response.py
```diff
@@ -0,0 +1,16 @@
...
+from ._create_asset_request import CreateAssetRequest
+
+
+class CreateAssetsPartialSuccessResponse(JsonModel):
+    """Model for create Assets Partial Success Response."""
+
+    assets: Optional[List[Asset]] = None
+    """Gets or sets array of created assets."""
+
+    failed: Optional[List[CreateAssetRequest]] = None
```
CiprianAnton, line 15: Are we missing errors field?

nisystemlink/clients/assetmanagement/models/_query_assets_request.py
```diff
@@ -0,0 +1,152 @@
...
+    take: Optional[int] = None
+    """Gets or sets how many resources to return in the result, or -1 to use a default defined by the service."""
+
+    order_by: Optional[OrderBy] = None
+    """Field by which assets can be ordered/sorted. If OrderBy is not specified, no sorting will applied."""
+
+    descending: Optional[bool] = None
+    """Whether to return the assets in the descending order. If OrderBy is not specified, this property is ignored."""
+
+    calibratable_only: Optional[bool] = None
```
CiprianAnton, line 105: I would not expose this since it only works when exporting as CSV

CiprianAnton: Same applies for ids

CiprianAnton (approved):

nisystemlink/clients/assetmanagement/models/_asset_create_request.py
aravindhan-ni: Updated

nisystemlink/clients/assetmanagement/_asset_management_client.py
```diff
@@ -0,0 +1,80 @@
...
+        """Create Assets.
+
+        Args:
+            assets: Array of assets that should be created.
+
+        Returns:
+            AssetsCreatePartialSuccessResponse: Array of created assets and array of failed.
+
+        Raises:
+            ApiException: If unable to communicate with the `nispec` service or if there are invalid
```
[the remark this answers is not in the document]
aravindhan-ni: Updated

nisystemlink/clients/assetmanagement/models/_query_assets_request.py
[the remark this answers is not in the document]
aravindhan-ni: Updated the projection field to use List[AssetField]

nisystemlink/clients/assetmanagement/models/_asset.py
aravindhan-ni: Updated as suggested

nisystemlink/clients/assetmanagement/models/_asset.py
aravindhan-ni: Updated as suggested

examples/assetmanagement/assets.py
aravindhan-ni: Updated to MANUAL

tests/integration/assetmanagement/test_asset_management.py
aravindhan-ni: referred and updated

tests/integration/assetmanagement/test_asset_management.py
aravindhan-ni: Added test

tests/integration/assetmanagement/test_asset_management.py
aravindhan-ni: Wrote a test to verify querying with specific IDs.

examples/assetmanagement/assets.py
aravindhan-ni: Updated

nisystemlink/clients/assetmanagement/models/_asset_location.py
aravindhan-ni: To add support to SLS, retaining these fields. [comment in another PR](https://github.com/ni/nisystemlink-clients-python/pull/86/files#r1949705749)

nisystemlink/clients/assetmanagement/models/_create_asset_request.py
aravindhan-ni: Added separate location class for create asset

examples/assetmanagement/assets.py
aravindhan-ni: Updated

nisystemlink/clients/assetmanagement/models/_create_assets_partial_success_response.py
rbell517: That comment only applied to the query response because that API will only use the error field when returning a failure status code. The create partial success will return a 200, which looks like a success and won't trigger the base class's error handling. You'll need to include the error field here as you had it before.

nisystemlink/clients/assetmanagement/models/_asset_calibration.py
```diff
@@ -0,0 +1,73 @@
...
+from enum import Enum
+from typing import List, Optional
+
+from nisystemlink.clients.core._uplink._json_model import JsonModel
+
+
+class CalibrationStatus(Enum):
+    """Calibration category the asset belongs to based on the next due calibration date."""
+
+    OK = "OK"
```
MuthuRajan-S, line 13: enum.auto()

Apply other places as well

nisystemlink/clients/assetmanagement/models/_query_assets_request.py
```diff
@@ -0,0 +1,152 @@
...
+    """
+    Gets or sets Whether to return the total number of assets which match the provided filter,
+    disregarding the take value.
+    """
+
+    filter: Optional[str] = None
+    """Gets or sets the filter criteria for assets. Consists of a string of queries composed using AND/OR operators."""
+
+
+class _QueryAssetRequest(JsonModel):
```
MuthuRajan-S, line 118: Is the underscore in the class name intentional?

tests/integration/assetmanagement/test_asset_management.py
```diff
@@ -0,0 +1,250 @@
...
+    QueryAssetsRequest,
+    QueryAssetsResponse,
+    SelfCalibration,
+    TemperatureSensor,
+)
+from nisystemlink.clients.core._http_configuration import HttpConfiguration
+
+
+@pytest.fixture(scope="class")
+def create_assets_request() -> List[CreateAssetRequest]:
```
MuthuRajan-S, line 27: Same as the other [comment](https://github.com/aravindhan-ni/nisystemlink-clients-python/pull/1#discussion_r2095794303).

MuthuRajan-S (approved):
Approving with minor comments

nisystemlink/clients/assetmanagement/models/_asset_location.py
```diff
@@ -0,0 +1,74 @@
...
+    slot_number: Optional[int] = None
+    """Gets or sets the number of the slot in which the asset is located."""
+
+    state: AssetPresenceWithSystemConnection
+    """Presence of an asset and the connection of the system in which it resides."""
+
+
+class AssetLocationForCreate(AssetLocation):
+
+    state: AssetPresence
```
rbell517, line 74: Include doc comments

nisystemlink/clients/assetmanagement/models/_create_asset_request.py
```diff
@@ -0,0 +1,98 @@
...
+    supports_external_calibration: Optional[bool] = None
+    """Gets or sets whether the asset supports external calibration."""
+
+    custom_calibration_interval: Optional[int] = None
+    """Gets or sets the interval represented in months used for computing calibration due date."""
+
+    self_calibration: Optional[SelfCalibration] = None
+    """Gets or sets the last self-calibration of the asset."""
+
+    is_NI_asset: Optional[bool] = None
```
rbell517, line 67: I expect you would need to set the field alias here like in the other case, please double check that this is not needed.

nisystemlink/clients/assetmanagement/models/_create_assets_partial_success_response.py
rbell517: The same applies to delete.

nisystemlink/clients/assetmanagement/models/_query_assets_request.py
```diff
@@ -0,0 +1,152 @@
...
+    OUT_FOR_CALIBRATION = "outForCalibration"
+
+
+class QueryAssetsRequest(JsonModel):
+    """Model for object containing filters to apply when retrieving assets."""
+
+    ids: Optional[List[str]] = None
+    """Gets or sets identifiers of the assets to be retrieved."""
+
+    projection: List[AssetField] = []
```
rbell517, line 87:
```suggestion
    projection: Optional[List[AssetField]] = None
```

nisystemlink/clients/assetmanagement/models/_query_assets_request.py
rbell517: Yes, see my comment in the client. We need an internal Pydantic model that uses the string project representation we convert to from the list of fields representation.

rbell517 (changes requested):

nisystemlink/clients/assetmanagement/models/_create_assets_partial_success_response.py
aravindhan-ni: Added API errors for create and delete APIs

tests/integration/assetmanagement/test_asset_management.py
```diff
@@ -0,0 +1,250 @@
...
+            ids=[asset_id], skip=0, take=1, returnCount=True
+        )
+
+        response: QueryAssetsResponse = client.query_assets(query=query_assets_request)
+
+        assert response is not None
+        assert response.assets is not None and len(response.assets) == 1
+        assert response.total_count >= 1
+
+    @responses.activate
```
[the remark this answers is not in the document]
aravindhan-ni: Updated

nisystemlink/clients/assetmanagement/_asset_management_client.py
```diff
@@ -0,0 +1,87 @@
...
+
+        Raises:
+            ApiException: If unable to communicate with the asset management service or if there are invalid
+            arguments.
+        """
+        ...
+
+    @post("query-assets")
+    def query_assets(
+        self, query: models.QueryAssetsRequest
```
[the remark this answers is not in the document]
aravindhan-ni: Updated as suggested

nisystemlink/clients/assetmanagement/models/_asset_calibration.py
```diff
@@ -0,0 +1,73 @@
...
+    """Gets or sets the manufacturer's recommended calibration interval in months."""
+
+    next_recommended_date: str
+    """Gets or sets ISO-8601 formatted timestamp specifying the recommended date for the next external calibration."""
+
+    next_custom_due_date: Optional[str] = None
+    """Gets or sets ISO-8601 formatted timestamp specifying the date for the next external calibration."""
+
+    resolved_due_date: Optional[str] = None
+    """Gets ISO-8601 formatted timestamp specifying the resolved due date for external calibration."""
```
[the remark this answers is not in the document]
aravindhan-ni: Updated

nisystemlink/clients/assetmanagement/models/_create_asset_request.py
aravindhan-ni: Added

nisystemlink/clients/assetmanagement/models/_query_assets_request.py
aravindhan-ni: Added

tests/integration/assetmanagement/test_asset_management.py
aravindhan-ni: Updated as suggested

nisystemlink/clients/assetmanagement/models/_asset_calibration.py
MuthuRajan-S: The only advantage of me recommending enum.auto() is to avoid code repetition, if its has issues, its not needed.

rbell517: If you used auto with a string enum instead that would be closer to what you want but it would be the lower case version of the string, which isn't what we want. I agree with dropping trying to use auto.

rbell517:
@MuthuRajan-S @aravindhan-ni I updated the repo settings so you can run the PR github actions without my approval.

nisystemlink/clients/assetmanagement/models/_asset_calibration.py
rbell517: You're still using auto in other enums. I've unresolved this comment chain to address that.

tests/integration/assetmanagement/test_asset_management.py
```diff
@@ -221,27 +180,27 @@ def test_query_assets_with_projections__returns_the_assets_with_projected_proper
             and asset.custom_calibration_interval is None
             and asset.discovery_type is None
             and asset.external_calibration is None
-            and asset.file_ids is None
+            and asset.file_ids == []
```
rbell517, line 183: When you're using a projection these should be None not empty lists.

nisystemlink/clients/assetmanagement/_asset_management_client.py
```diff
@@ -71,6 +71,44 @@ def query_assets(
...
+            "order_by": query.order_by,
+            "descending": query.descending,
+            "projection": projection_str,
+        }
+
+        query_params = {k: v for k, v in query_params.items() if v is not None}
+
+        query_request = models._QueryAssetRequest(**query_params)
+
+        print(query_request, projection_str, query.projection)
```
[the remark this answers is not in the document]
aravindhan-ni: Removed

nisystemlink/clients/assetmanagement/models/_asset_calibration.py
aravindhan-ni: Updated

tests/integration/assetmanagement/test_asset_management.py
rbell517: That sounds like a bug in the service's projection implementation. @CiprianAnton is this the behavior you expect?

aravindhan-ni:
@rbell517 Please go ahead and merge the PR if there are no further comments or concerns.

PR #147, opened 2025-06-12
feat: Spec client | Add Get Spec API 

vishnu-ni:

This PR adds the `Get spec` API of the `Specification management service` to the spec client module.

It is helpful for users to retrieve a specification directly by its ID, rather than using the query API to fetch a single specification.

API link: [Swagger link](https://dev-api.lifecyclesolutions.ni.com/niapis/?urls.primaryName=Specification+Management#/Specifications/get_nispec_v1_specs__id_)

### DIFF
--- a/nisystemlink/clients/core/_api_error_response.py
+++ b/nisystemlink/clients/core/_api_error_response.py
@@ -1,7 +1,8 @@
 from nisystemlink.clients.core._api_error import ApiError
+from nisystemlink.clients.core._uplink._json_model import JsonModel
 
 
-class ApiErrorResponse:
+class ApiErrorResponse(JsonModel):
     """Represents an error response from the SystemLink API responses."""
 
     error: ApiError
--- a/nisystemlink/clients/core/_api_error_response.py
+++ b/nisystemlink/clients/core/_api_error_response.py
@@ -1,8 +1,0 @@
-from nisystemlink.clients.core._api_error import ApiError
-from nisystemlink.clients.core._uplink._json_model import JsonModel
-
-
-class ApiErrorResponse(JsonModel):
-    """Represents an error response from the SystemLink APIs."""
-
-    error: ApiError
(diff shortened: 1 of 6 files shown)


nisystemlink/clients/spec/_spec_client.py
```diff
@@ -96,6 +97,22 @@ def query_specs(
         """
         ...
 
+    @get("specs/{id}")
+    def get_spec(self, id: str) -> Union[models.Specification, ApiErrorResponse]:
```
vishnu-ni, line 101: Here we can use `models.specification | ApiErrorResponse` but the pipe ` | ` is only supported in Python version greater than 3.10. That's why we are using `Union` here.

nisystemlink/clients/core/_api_error_response.py
```diff
@@ -0,0 +1,8 @@
+from nisystemlink.clients.core._api_error import ApiError
+from nisystemlink.clients.core._uplink._json_model import JsonModel
+
+
+class ApiErrorResponse(JsonModel):
```
vishnu-ni, line 5: The reason for making this class a Pydantic model is that it is used as a response model for the get_spec API. Since Uplink relies on Pydantic to generate schemas for request and response models, the response class must inherit from a Pydantic BaseModel (or a subclass like JsonModel) to ensure compatibility and avoid schema generation errors.

hariram-ni (approved):

MuthuRajan-S (approved):

tests/integration/spec/test_spec.py
```diff
@@ -237,6 +239,47 @@ def test__update_single_same_version__version_updates(
         updated_spec = update_response.updated_specs[0]
         assert updated_spec.version == 1
 
+    def test__get_spec_by_id__spec_matches_expected(
+        self, client: SpecClient, create_specs, product
+    ):
+        productId = product
```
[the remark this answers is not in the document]
vishnu-ni: I think we can leave it as is, since we are also using this variable in one of the [assertion](https://github.com/ni/nisystemlink-clients-python/pull/147/files#:~:text=.product_id%20%3D%3D-,productId,-def%20test__get_non_existant_spec_by_id__get_spec_fails()). The variable name `productId` clearly indicates its purpose, whereas using `product` could be confusing, as it sounds like it refers to the entire product rather than just the ID.

PR #150, opened 2025-07-21
fix: upgrade uplink to v0.10.0 for python 3.10+ to resolve pkg_resources deprecation

### LINKED ISSUE #91: Uplink (base for clients) seems to be abandoned
Based on [this issue](https://github.com/prkumar/uplink/issues/311) and the fact that the last release was in 2022, it would appear that uplink has been abandoned. It has 64 open issues, and while I don't immediately see anything security related, that scares me given that there are only 63 closed ones. (Also, issue our #77 seems to be a direct result of [this issue.](https://github.com/prkumar/uplink/issues/305) )

It seems to me like it would be worth refactoring to remove uplink or forking it and building some meaningful knowledge around it.

fredvisser:

The latest version of `uplink` resolves a deprecated package warning, but also requires Python 3.10+. This PR conditionally upgrades `uplink` for applications that support 3.10+.

- Upgrades uplink from 0.9.7 to 0.10.0 with pydantic extras for Python 3.10+, but maintains uplink 0.9.7 for Python 3.9.
- Resolves pkg_resources deprecation warning by using importlib.metadata

Fixes #91 

Every time the `nisystemlink-client` library is called, the following warning is printed to the console:

```
[...]python3.11/site-packages/uplink/_extras.py:4: UserWarning: pkg_resources is deprecated as an API. See 
https://setuptools.pypa.io/en/latest/pkg_resources.html. The pkg_resources package is slated for removal as 
early as 2025-11-30. Refrain from using this package or pin to Setuptools<81.
```

Uplink 0.10.0 was released in June to resolve this dependency issue, but requires the use of Python 3.10+.

- All tests pass and code quality checks pass


pyproject.toml
```diff
@@ -31,12 +30,12 @@ classifiers = [
 ]
 
 [tool.poetry.dependencies]
-python   = "^3.9"
+python   = "^3.10"
```
rbell517, line 33: NI's python library policy is to always support all non-EOL python versions, which includes 3.9 until October 31 of this year.

What we actually want in this case is to conditionally require the higher version when installing to python 3.10 or newer. Poetry [supports this](https://python-poetry.org/docs/dependency-specification/#multiple-constraints-dependencies).

```toml
uplink = [
  { version = "^0.10.0", extras = ["pydantic"], python_version = ">=3.10" },
  { version = "^0.9.1", python_version = ">=3.9,<3.10" }
]
```

rbell517 (changes requested):

PR #84, opened 2024-11-11
feat: Add pandas utility for dataframe

ancy-augustin:
- [ ] This contribution adheres to [CONTRIBUTING.md](https://github.com/ni/nisystemlink-clients-python/blob/master/CONTRIBUTING.md).

TODO: Check the above box with an 'x' indicating you've read and followed [CONTRIBUTING.md](https://github.com/ni/nisystemlink-clients-python/blob/master/CONTRIBUTING.md).

### What does this Pull Request accomplish?

Adds utility functions to interact with SystemLink `DataFrame` API using pandas data frames. 

### Why should this Pull Request be merged?

Simplifies working with pandas data frames and SystemLink `DataFrame` by providing convenient methods for creating, appending, and querying tables directly from pandas.

### What testing has been done?

TODO: Detail what testing has been done to ensure this submission meets requirements.

PR #155, opened 2025-09-11
feat: DataFrame client binary ingestion support

cameronwaterman:

DFS append table data now supports binary ingestion via Apache Arrow.

Increase feature set of DFS API.

Added integration tests.


nisystemlink/clients/dataframe/_data_frame_client.py
```diff
@@ -250,18 +253,160 @@ def get_table_data(
...
+        args=[Path, Body, Query("endOfData")],
+        content_type="application/vnd.apache.arrow.stream",
+    )
+    def _append_table_data_arrow(
+        self, id: str, data: bytes, end_of_data: Optional[bool] = None
+    ) -> None:
+        """Internal uplink-implemented Arrow (binary) append call."""
+        ...
+
+    def append_table_data(
```
adamarnesen, line 273: In #84, we're trying to add APIs that take pandas frames as input and do the formatting manipulation for the user. Is that something we should consider here too? Just wondering how we want to merge those.

spanglerco: Well, it's generally trivial to go from Pandas to Arrow:

```py
import pyarrow as pa

table = pa.Table.from_pandas(df)
client.append_table_data(table.to_batches(max_chunksize=<optional>))
```

Note that we explicitly ruled out Pandas as first-class in [the HLD](https://dev.azure.com/ni/DevCentral/_wiki/wikis/Stratus/140836/Binary-ingestion-API#alternative-6%3A-support-additional-data-parameter-types-in-the-python-api).

rbell517 (approved):

nisystemlink/clients/dataframe/_data_frame_client.py
```diff
@@ -250,18 +256,181 @@ def get_table_data(
...
+                    "pyarrow is not installed. Install to stream RecordBatches."
+                )
+
+            if not isinstance(first_batch, pa.RecordBatch):
+                raise ValueError(
+                    "Iterable provided to data must yield pyarrow.RecordBatch objects."
+                )
+
+            def _generate_body() -> Iterable[memoryview]:
+                data_iter = iter(data)
```
spanglerco, line 362: This will iterate a second time, which may lead to unexpected behavior if clients are generating their record batches. We should be able to pass in the `iterator` and `first_batch` from above and use those.

nisystemlink/clients/dataframe/_data_frame_client.py
```diff
@@ -250,18 +256,181 @@ def get_table_data(
...
+                            yield slice
+                        buf.seek(0)
+                        try:
+                            batch = next(data_iter)
+                        except StopIteration:
+                            break
+
+                    writer.close()
+                    with buf.getbuffer() as view:
+                        with view[0 : buf.tell()] as slice:
```
spanglerco, line 383: nit: This part uses nested `with`s, while line 373 combines them into one. Can they both use the same style?

tests/integration/dataframe/test_dataframe.py
```diff
@@ -615,3 +628,217 @@ def test__export_table_data__works(self, client: DataFrameClient, create_table):
                 response.read()
                 == b'"col1","col2","col3"\r\n1,2.5,6.5\r\n2,1.5,5.5\r\n3,2.5,7.5'
             )
+
+    def test__append_table_data__append_request_success(
```
spanglerco, line 632: Can we have these tests read back the table metadata and verify the `RowCount` and `SupportsAppend` are correct?

tests/integration/dataframe/test_dataframe.py
```diff
@@ -615,3 +628,217 @@ def test__export_table_data__works(self, client: DataFrameClient, create_table):
...
+            ValueError, match="end_of_data must be provided when data is None"
+        ):
+            client.append_table_data(table_id, None)
+
+    def test__append_table_data__flush_only_with_none(
+        self, client: DataFrameClient, create_table
+    ):
+        table_id = self._new_single_int_table(create_table)
+        frame = DataFrame(columns=["a"], data=[["1"]])
+        client.append_table_data(table_id, frame)
```
spanglerco, line 688: nit: Appending data before setting end of data isn't required. The test could verify `RowCount` is 0 and `SupportsAppend` is false.

tests/integration/dataframe/test_dataframe.py
```diff
@@ -615,3 +628,217 @@ def test__export_table_data__works(self, client: DataFrameClient, create_table):
...
+
+        assert "Arrow ingestion request was rejected" not in str(excinfo.value)
+
+    def test__append_table_data__arrow_ingestion_non_400_passthrough(
+        self, client: DataFrameClient
+    ):
+        pa = pytest.importorskip("pyarrow")
+        table_id = "mock_table_id"
+        batch = pa.record_batch([pa.array([1, 2, 3])], names=["a"])
+        with responses.RequestsMock() as rsps:
```
spanglerco, line 828: I don't feel strongly about it, but you could easily get a 404 by passing a (valid) table ID that doesn't exist.

tests/integration/dataframe/test_dataframe.py
```diff
@@ -615,3 +628,217 @@ def test__export_table_data__works(self, client: DataFrameClient, create_table):
...
+            with pytest.raises(ApiException) as excinfo:
+                client.append_table_data(table_id, [batch], end_of_data=True)
+        assert "Arrow ingestion request was rejected" not in str(excinfo.value)
+
+    def test__append_table_data__unsupported_type_raises(
+        self, client: DataFrameClient, create_table
+    ):
+        table_id = create_table(basic_table_model)
+        with pytest.raises(ValueError, match="Unsupported type"):
+            # cast to Any to satisfy type checker while still exercising runtime path
```
spanglerco, line 843: Does `# type: ignore` work? Might be more obvious than the cast.

spanglerco (changes requested):

tests/tag/test_tagmanager.py
```diff
@@ -2383,12 +2386,13 @@ def test__create_writer_with_buffer_time__sends_when_timer_elapsed(self):
 
         writer.write(path, tbase.DataType.INT32, value, timestamp=timestamp)
         self._client.all_requests.assert_not_called()
-        for i in range(100):
-            if self._client.all_requests.call_count > 0:
-                break
-            time.sleep(0.01)
-
-        utctime = datetime.utcfromtimestamp(timestamp.timestamp()).isoformat() + "Z"
+        # Deterministic flush instead of waiting for timer.
```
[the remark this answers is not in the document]
cameronwaterman: 🤦‍♂️it would help if I actually paid attention huh

tests/integration/dataframe/test_dataframe.py
spanglerco: I intended this comment to apply to all of the tests that are only calling `append_table_data` without any asserts, not just `test__append_table_data__flush_only_with_none`.

examples/dataframe/create_write_data.py
```diff
@@ -25,12 +29,48 @@
...
+            names=["ix", "Float_Column", "Timestamp_Column"],
+        )
+    ]
+    client.append_table_data(table_id, batch_list)
+
+    # Mark end_of_data for the table
+    # Supply `None` and `end_of_data=True`
+    client.append_table_data(table_id, None, end_of_data=True)
+    # OR supply an empty Iterable and `end_of_data=True`
+    client.append_table_data(table_id, [], end_of_data=True)
```
spanglerco, line 73: It looks like my previous comment about not needing to explicitly show using an empty iterable got lost. I think it's fine to show that using None is the correct way to set end of data without having data to append. An empty iterable will "just work", but not recommended to do on purpose.

```suggestion
```

spanglerco (approved):

PR #164, opened 2025-10-20
feat: DFS support for test result ID, column name queries, and interactive queries

spanglerco:

Add a `test_result_id` field to all metadata models (create, read, and modify). Add an `interactive` field to the query tables request model. Document the ability to filter tables by test result ID and column names.

The DFS client should support new functionality being added to the service.

Ran the DFS tests locally against:

1. The test cluster that has the new functionality
2. The demo cluster that has the functionality disabled (feature toggle is off)
3. A local build of the 2025-10 version of the service (before the functionality was added)

Verified that all tests passed. Cases 2 and 3 log a warning that the service doesn't support test result IDs.
Verified that the tests pass in the GitHub workflow without logging the warning about test result ID support.
Verified the built docs look correct, including the reference to the various operations.

### DIFF
--- a/tests/integration/dataframe/test_dataframe.py
+++ b/tests/integration/dataframe/test_dataframe.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 import warnings
 from datetime import datetime, timezone
-from typing import Callable, List, Optional
+from typing import Callable, List, Optional, TypedDict
 
 import pytest  # type: ignore
 import responses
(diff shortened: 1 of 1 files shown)


nisystemlink/clients/dataframe/models/_modify_table_request.py
```diff
@@ -11,8 +11,7 @@ class ColumnMetadataPatch(JsonModel):
 
     properties: Dict[str, Optional[str]]
     """The properties to modify. A map of key value properties containing the metadata
-    to be added or modified. Setting a property value to ``None`` will delete the
-    property."""
+    to be added or modified. Set a property value to ``None`` to delete the property."""
```
spanglerco: Note: Updated some of the wording to match changes to the DFS OpenAPI document.

PR #166, opened 2025-10-22
fix: Fix typing in examples and tests

spanglerco:

Make mypy checking more strict for Pydantic models and fix or suppress the resulting errors.

I noticed that the DFS create_write_data example gets Pydantic validation errors at run-time because it passes numeric values instead of converting them to strings. This led me to discover that Pydantic's mypy plugin causes typing to be more permissive by default. Configuring it to be more strict resulted in 58 type errors, mostly typos that get normalized to the correct camelCase or using strings when other types were expected.

Verified all of the DFS examples work. I ran the Test Monitor steps example and it gets farther than it did before my change, but gets a 400 Bad Request. I intend to open a separate issue for that and an issue in the Asset Management assets example.

I temporarily set `init_forbid_extra = True` in the mypy.ini, which revealed a few other issues. However, the Test Monitor client uses extra fields, so I didn't leave that setting on. I suppose we could leave it on and add `# type: ignore` to the places we use extra fields, but I didn't try that.

### DIFF
--- a/examples/test_plan/test_plans.py
+++ b/examples/test_plan/test_plans.py
@@ -35,7 +35,7 @@ create_test_plans_request = [
         dut_serial_number="serial_number_123",
         test_program="TP-Integration-001",
         system_filter="os:linux AND arch:x64",
-        dut_Filter="modelName = 'cRIO-9045' AND serialNumber = '01E82ED0'",
+        dut_filter="modelName = 'cRIO-9045' AND serialNumber = '01E82ED0'",
         workspace="your_workspace_id",
         file_ids_from_template=["file1", "file2"],
         dashboard=Dashboard(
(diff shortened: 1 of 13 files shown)


examples/test_plan/test_plans.py
```diff
@@ -69,7 +69,7 @@
 
 # Query test plan using id.
 query_test_plans_request = QueryTestPlansRequest(
-    skip=0, take=1, descending=False, returnCount=False
+    take=1, descending=False, return_count=False
```
spanglerco: Note: `QueryTestPlansRequest` doesn't have a `skip` property (though it does have a `continuation_token` property). Since it was passing `skip=0`, I'm assuming it would have no effect whether the service supports a `skip` parameter or not.

examples/testmonitor/steps.py
```diff
@@ -44,7 +46,7 @@ def create_test_result():
 create_response = create_test_result()
 
 # Create the step requests
-result_id = create_response.results[0].result_id
+result_id = cast(str, create_response.results[0].id)
```
spanglerco: Note: The `Result` model doesn't have a `result_id` property, but it does have an `id` property that is marked Optional (presumably because of projections).

rbell517 (approved):

PR #177, opened 2025-12-10
feat: Add optional take to DataFrame Service export table data request model

spanglerco:

Python client support for restricting the number of rows returned by the export table data route. The default is to return all rows.

Enables using the export route as an undecimated dump of table data while restricting the number of rows to avoid retrieving more data than expected.

Tests pass. Ran the export_data example against a table with over 100,000 rows and verified the resulting CSV file only contained 100,000 rows.

### DIFF
--- a/tests/integration/dataframe/test_dataframe.py
+++ b/tests/integration/dataframe/test_dataframe.py
@@ -309,23 +309,6 @@ class TestDataFrame:
 
         return (column_names, rows)
 
-    @staticmethod
-    def _create_expected_csv(columns: List[Column], data: _DataFrameData) -> str:
-        output = io.StringIO()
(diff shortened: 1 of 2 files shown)


tests/integration/dataframe/test_dataframe.py
```diff
@@ -105,6 +121,26 @@ def supports_test_result_id(client: DataFrameClient) -> bool:
     return result
 
 
+@pytest.fixture(scope="class")
+def table_with_data(client: DataFrameClient, create_table) -> str:
```
spanglerco: Note: I further adjusted the tests to only create a table with data once and reuse that table across multiple tests. That way we don't have to wait for the data to be available multiple times.

rbell517 (approved):

PR #185, opened 2025-12-24
feat: Pagination helper

santhoshramaraj:

Simplify the use of APIs with Paged Responses

Unit tests with mock

🪄Developed with AI help

### DIFF
--- a/tests/core/test_pagination.py
+++ b/tests/core/test_pagination.py
@@ -233,7 +233,9 @@ class TestPaginate:
         mock_fetch = MagicMock(
             side_effect=[
                 MockResponseWithItems([1, 2], "token1"),
-                MockResponseWithItems([3, 4], "token1"),  # Same token - should raise error
+                MockResponseWithItems(
+                    [3, 4], "token1"
+                ),  # Same token - should raise error
             ]
         )
 


nisystemlink/clients/core/helpers/_pagination.py
```diff
@@ -0,0 +1,59 @@
+# -*- coding: utf-8 -*-
+from typing import Any, Callable, Generator, TypeVar
+
+ItemType = TypeVar("ItemType")
+
+
+def paginate(
+    fetch_function: Callable[..., Any],
+    items_field: str = "items",
```
spanglerco: Should we make this required? I don't see any models that derive from `WithPaging` that have an `items` attribute, so callers will always need to specify this.
```suggestion
    items_field: str,
```

nisystemlink/clients/core/helpers/_pagination.py
```diff
@@ -0,0 +1,59 @@
...
+
+        # Yield each item individually
+        for item in items:
+            yield item
+
+        # Get the continuation token for the next page
+        continuation_token = getattr(response, continuation_token_field, None)
+
+        # Stop if there are no more pages
+        if continuation_token is None:
```
spanglerco: Should we guard against the continuation token not changing between iterations, leading to an infinite loop? For example, we could raise an exception since, most likely, it's a usage error or bug.

spanglerco (changes requested):

tests/core/test_pagination.py
```diff
@@ -0,0 +1,245 @@
...
+
+class MockResponse:
+    """Mock API response object for testing pagination."""
+
+    def __init__(self, items: List[Any], continuation_token: str | None = None):
+        self.items = items
+        self.continuation_token = continuation_token
+
+
+class MockResponseCustomFields:
```
spanglerco: Does it still make sense to have both `MockResponse` and `MockResponseCustomFields` now that `items_field` is required? I suppose it's still good to test with different values. Maybe we should rename these classes instead, like `MockResponseWithItems` and `MockResponseWithResults`?

PR #195, opened 2026-01-28
feat: Add Notification Client

IN2026-Rohith-Raja:
Notification Client - Apply Dynamic Strategy

Created Notification Client implementation for the API `/ninotification/v1/apply-dynamic-strategy`. This includes various models required for the request body and performed various tests.

Used pytest to test the client working. Includes tests that covers negative cases like invalid configurations, invalid recipient address, invalid message template and other cases.

### DIFF
--- a/nisystemlink/clients/notification/models/_smtp_address_group.py
+++ b/nisystemlink/clients/notification/models/_smtp_address_group.py
@@ -44,7 +44,9 @@ class SmtpAddressGroup(BaseNotificationMetadata):
 
     @model_validator(mode="before")
     @classmethod
-    def set_interpreting_service_name(cls, data: "SmtpAddressGroup") -> "SmtpAddressGroup":
+    def set_interpreting_service_name(
+        cls, data: "SmtpAddressGroup"
+    ) -> "SmtpAddressGroup":
         if isinstance(data, dict) and "interpreting_service_name" not in data:
             data["interpreting_service_name"] = "smtp"
         return data


nisystemlink/clients/notification/models/_common_meta_data.py
SSadaiyappan-NI: rename this file to same name as the class

nisystemlink/clients/notification/models/_address_group.py
```diff
@@ -0,0 +1,44 @@
...
+
+    bccAddresses: List[str] | None = None
+    """List of blind carbon copy recipient addresses."""
+
+
+class AddressGroup(BaseNotificationMetadata):
+    """Model defining notification recipients."""
+
+    interpreting_service_name: str
+    """Gets or sets the name of the interpreting service.
```
rbell517: I know we're just lifting this from the swagger, which is generated from the C# classes, but many of these are not very useful and we should not be using this "Gets or sets" language.

rbell517 (changes requested):

tests/integration/notification/test_notification_client.py
```diff
@@ -0,0 +1,229 @@
...
+    """Returns the created address group."""
+    return SmtpAddressGroup(
+        id="address_group_id",
+        interpreting_service_name="smtp",
+        display_name="name",
+        properties={"property": "value"},
+        fields=SmtpAddressFields(
+            toAddresses=["address1@example.com"],
+            ccAddresses=["address2@example.com"],
+            bccAddresses=["address3@example.com"],
```
rbell517: We don't want to actually send emails to these email addresses when these tests run. We should limit test side effects to services we control on the test tier. In cases like this we should fake the actual HTTP requests sent. You can see the [notebook client tests](https://github.com/ni/nisystemlink-clients-python/blob/master/tests/integration/notebook/test_notebook_client.py) for examples of how to do this.

nisystemlink/clients/notification/models/_smtp_address_group.py
```diff
@@ -0,0 +1,42 @@
...
+    """List of carbon copy recipient addresses."""
+
+    bccAddresses: List[str] | None = None
+    """List of blind carbon copy recipient addresses."""
+
+
+class SmtpAddressGroup(BaseNotificationMetadata):
+    """Model defining notification recipients for SMTP service."""
+
+    interpreting_service_name: Literal["smtp"]
```
rbell517: If they're using this class we don't need to also ask them to fill in this field. Same for the message template.

```suggestion
    interpreting_service_name: Literal["smtp"] = "smtp"
```

tests/integration/notification/test_notification_client.py
```diff
@@ -0,0 +1,248 @@
...
+    SmtpMessageTemplate,
+    SmtpMessageTemplateFields,
+)
+from pydantic import ValidationError
+
+BASE_URL = "https://test-api.lifecyclesolutions.ni.com"
+
+
+@pytest.fixture
+def _address_group():
```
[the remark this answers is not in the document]
SSadaiyappan-NI: Now that we are only testing smtp service, see if we could rename the tests to convey it

tests/integration/notification/test_notification_client.py
SSadaiyappan-NI: Add tests to verify the new validator you have created.
1. Address group or Address group id must be present
2. Message template or message template id must be present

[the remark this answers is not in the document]
IN2026-Rohith-Raja: Used model validator to set the service name.

PR #210, opened 2026-05-26
Make alias handling consistent and clean up lint findings

fredvisser:
## Summary
- make `JsonModel` alias handling accept Python snake_case while preserving API/JSON field name compatibility
- convert explicit alias fields to the same pattern and rename the asset Python field to `is_ni_asset` while keeping legacy aliases accepted
- add regression coverage for alias behavior and clean up the remaining docstring, naming, and lint findings in touched files

## Why

Part of the motivation for this change is editor and AI/tooling behavior.

Previously, Pydantic alias handling caused camelCase API/JSON names to appear in IDE-visible constructor signatures. That meant editor completions and AI suggestions would often prefer `programName=` over the Python field name `program_name=`.

This PR makes Python field names primary in the constructor surface while preserving API/JSON compatibility:
- Python code sees and uses snake_case names
- serialization still emits camelCase API/JSON names
- legacy camelCase inputs are still accepted during validation

That keeps the wire format unchanged while making the Python API surface more consistent for developers and tools.

## Validation
- `poetry run flake8 nisystemlink tests`
- `poetry run black --check nisystemlink examples tests`
- `poetry run pytest`
- `poetry run pytest -q tests/core/test_json_model.py`
- `poetry run pytest --collect-only -q tests/integration/assetmanagement/test_asset_management.py`

### DIFF
--- a/nisystemlink/clients/assetmanagement/models/_asset.py
+++ b/nisystemlink/clients/assetmanagement/models/_asset.py
@@ -2,7 +2,7 @@ from datetime import datetime
 from typing import Dict, List
 
 from nisystemlink.clients.core._uplink._json_model import JsonModel
-from pydantic import Field
+from pydantic import AliasChoices, Field
 
 from ._asset_calibration import (
     CalibrationStatus,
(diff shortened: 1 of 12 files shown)


rbell517:
FYI notebook client test thats failing in your build is a known intermittent test that I believe is actually a backend bug that I haven't been able to track down. For now I suggest just re-running the failed run.

rbell517:
Also could you make a note in your PR description about how this impacts editor or AI behavior, if I remember right that was part of the motivation for this change.

rbell517 (approved):

PR #211, opened 2026-05-28
Add single-item convenience wrappers for bulk partial-success APIs

fredvisser:
## Summary

Several SystemLink client APIs only expose bulk operations that return partial-success envelopes (`items` + `failed` + `error`). When callers need to create or update a **single** item — by far the most common case — they must wrap the call in a list, then manually unpack the response and check for partial failures. This PR eliminates that boilerplate by adding singular convenience methods to six client classes.

## What changed

| Client | New methods |
|--------|------------|
| `TestMonitorClient` | `create_result`, `create_step`, `update_result`, `update_step` |
| `ProductClient` | `create_product`, `update_product` |
| `SpecClient` | `create_spec`, `update_spec` |
| `TestPlanClient` | `create_test_plan`, `update_test_plan`, `schedule_test_plan`, `create_test_plan_template` |
| `WorkItemClient` | `create_work_item`, `update_work_item`, `schedule_work_item`, `create_work_item_template`, `update_work_item_template` |
| `AssetManagementClient` | `create_asset` |

A shared helper `unwrap_single_item_partial_success()` (in `nisystemlink/clients/core/helpers/_partial_success.py`) centralizes the unwrap-or-raise logic so each wrapper stays minimal.

## Why this is beneficial

1. **Reduces caller boilerplate** — common single-item operations go from ~6 lines to 1.
2. **Safer error handling** — partial-success failures are translated into `ApiException` with full error context, rather than silently returning empty lists.
3. **Consistent pattern** — all wrappers follow the same contract (raise on failure, return typed item on success), making the API surface predictable.
4. **Non-breaking** — the existing bulk methods remain unchanged; the new methods are purely additive.

## Testing

- 24 new unit tests covering the shared helper and all 18 convenience methods.
- Full CI validation passes: `poe check` (formatting), `poe lint`, `poe types` (mypy, 379 source files), `poe test` (410 tests passed).

### DIFF
--- a/nisystemlink/clients/assetmanagement/_asset_management_client.py
+++ b/nisystemlink/clients/assetmanagement/_asset_management_client.py
@@ -13,6 +13,7 @@ from nisystemlink.clients.core._uplink._methods import post
 from uplink import Field, Path, retry
 
 from . import models
+from ..core.helpers._partial_success import unwrap_single_item_partial_success
 
 
 @retry(
@@ -55,6 +56,30 @@ class AssetManagementClient(BaseClient):
             arguments.
         """
         ...
+
+    def create_asset(self, asset: models.CreateAssetRequest) -> models.Asset:
+        """Create a single asset.
+
+        Args:
+            asset: The asset to create.
+
+        Returns:
+            The created asset.
+
+        Raises:
+            ApiException: if the asset could not be created or the service returns an
+                unexpected partial-success payload.
+        """
+        response = self.create_assets([asset])
+
+        return unwrap_single_item_partial_success(
+            response=response,
+            items=response.assets,
+            failed=response.failed,
+            error=response.error,
+            failure_message="Failed to create asset.",
+            empty_message="Server returned no created assets.",
+        )
 
     @post("query-assets")
     def __query_assets(
--- a/nisystemlink/clients/core/helpers/_partial_success.py
+++ b/nisystemlink/clients/core/helpers/_partial_success.py
@@ -1,0 +1,41 @@
+from typing import Any, Sequence, TypeVar
+
+from nisystemlink.clients import core
+
+_ItemT = TypeVar("_ItemT")
+
+
+def unwrap_single_item_partial_success(
+    *,
+    response: Any | None,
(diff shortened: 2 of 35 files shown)


fredvisser:
Addressed the review feedback in 4ca9b25. The shared single-item partial-success helper now raises ApiException when the service returns anything other than exactly one successful item, and I added a regression test for the multi-item case. Focused validation: pytest tests/core/test_partial_success.py.

nisystemlink/clients/core/helpers/_partial_success.py
```diff
@@ -0,0 +1,47 @@
...
+    response_data = (
+        response.model_dump(mode="json", by_alias=True)
+        if response is not None
+        else None
+    )
+
+    if failed or error:
+        raise core.ApiException(
+            failure_message,
+            error=error,
```
rbell517: The error here in the case of partial success on a single item request is very likely to be a OneOrMoreErrorsOccurred error with a single InnerErrors error. That inner error is the actually useful part to a caller.

What you have to leave it as a one or more isn't wrong and it would be fine to submit like you have it. A nice to have in this unwrapping would be to also unwrap this one or many into just one error, if there is really only one inner error. You can identify one of these by name (`Skyline.OneOrMoreErrorsOccurred`) or code (`-251041`) the check inner_errors for length one.

nisystemlink/clients/assetmanagement/_asset_management_client.py
```diff
@@ -56,6 +57,30 @@ def create_assets(
         """
         ...
 
+    def create_asset(self, asset: models.CreateAssetRequest) -> models.Asset:
```
rbell517, line 60: Your PR description claims that single item create is expected to be the most common. If so we should update the examples to use the new single item convenience functions where applicable. [Assets create example](https://github.com/ni/nisystemlink-clients-python/blob/master/examples/assetmanagement/assets.py) is a good case for updating.

rbell517: I would be fine with this as a follow up PR as well.

rbell517 (approved):

PR #213, opened 2026-06-05
Add explicit service package exports

fredvisser:
## Summary
- add short "Start here" package docstrings that point readers to the primary service client or tag entry point
- replace blanket `# flake8: noqa` package exports with explicit `__all__` declarations across service packages
- add a regression test that verifies each service package documents and declares its public exports

## Why This Helps
- it turns the package surface into an explicit contract instead of leaving it implicit in import side effects or wildcard exports
- it gives IDEs, static analysis, and coding agents a reliable list of supported imports, which reduces guesswork about whether a symbol is public or internal
- the new "Start here" docstrings give both humans and agents a canonical entry point, which makes generated examples and import suggestions more likely to use the intended top-level client
- the export test keeps that contract from drifting, so future refactors are less likely to break discoverability or accidentally hide supported symbols

## Validation
- `poetry run poe check`
- `poetry run poe lint`
- `poetry run poe types`
- `poetry run poe test`

## Notes
- `nisystemlink.clients.tag` keeps a few narrow `# noqa: I100` import-order suppressions because its runtime-safe import order is intentionally different from alphabetical order; the touched tag import path is covered by tests

### DIFF
--- a/nisystemlink/clients/alarm/__init__.py
+++ b/nisystemlink/clients/alarm/__init__.py
@@ -1,3 +1,5 @@
+"""Start here with AlarmClient for alarm management."""
+
 from nisystemlink.clients.alarm._alarm_client import AlarmClient
 
-# flake8: noqa
+__all__ = ["AlarmClient"]
--- a/nisystemlink/clients/artifact/__init__.py
+++ b/nisystemlink/clients/artifact/__init__.py
@@ -1,3 +1,5 @@
+"""Start here with ArtifactClient for artifact management."""
+
 from ._artifact_client import ArtifactClient
 
-# flake8: noqa
+__all__ = ["ArtifactClient"]
--- a/nisystemlink/clients/assetmanagement/__init__.py
+++ b/nisystemlink/clients/assetmanagement/__init__.py
@@ -1,5 +1,7 @@
+"""Start here with AssetManagementClient for asset management."""
+
 from nisystemlink.clients.assetmanagement._asset_management_client import (
     AssetManagementClient,
 )
 
-# flake8: noqa
+__all__ = ["AssetManagementClient"]
--- a/nisystemlink/clients/dataframe/__init__.py
+++ b/nisystemlink/clients/dataframe/__init__.py
@@ -1,3 +1,5 @@
+"""Start here with DataFrameClient for dataframe operations."""
+
 from ._data_frame_client import DataFrameClient
 
-# flake8: noqa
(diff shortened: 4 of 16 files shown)


nisystemlink/clients/alarm/__init__.py
```diff
@@ -1,3 +1,5 @@
+"""Start here with AlarmClient for alarm management."""
+
 from nisystemlink.clients.alarm._alarm_client import AlarmClient
 
-# flake8: noqa
+__all__ = ["AlarmClient"]
```
rbell517, line 5: Can you comment on whether we get similar benefits from doing the same for each client's models `__init__` as well? Doesn't need to be done now, it would just help me understand the impact of this.

rbell517 (approved):

nisystemlink/clients/alarm/__init__.py
fredvisser: Short answer: yes - that looks like a useful follow-up change 👍

What would improve:
1. Better public contract clarity
Today most models package initializers in `__init__.py`, `__init__.py`, `__init__.py`, etc. end with broad `flake8` suppression and rely on imported names being implicitly public. Explicit export lists would make that contract intentional and reviewable, same as we did for service package initializers.

2. Better tooling and agent behavior
A lot of code in this repo imports from models package initializers (examples, tests, and library internals), for example in notebook.py, test_alarm_client.py, and _dataframe_utilities.py. Explicit exports would give IDEs and coding agents a cleaner signal for what is intended to be imported from the package root versus internal module files.

3. Stronger regression protection
If we add a companion test (similar in spirit to the new service export test), we can catch accidental surface drift when models are added/removed/renamed.

Drawbacks:
1. Scope and risk
Models initializers are much larger and more numerous than service initializers, so changing all of them at once is a materially bigger compatibility surface.

2. Maintenance overhead
Many model files are generated or churn frequently with API changes. Hand-maintained export lists across all models packages could become noisy unless we automate generation/verification.

3. No single start-here concept
For service packages there is a clear primary entry point (client class). For models packages there is usually a broad set of peer types, so a start-here docstring is less obvious and provides less guidance value.

I'll make a follow-up PR that:
1. Is focused only on models package exports.
2. Keeps behavior non-breaking: preserve current exported names, just make them explicit.
3. Adds one repo-wide test to assert declared exports match imported package-surface symbols for models initializers.
4. And, if possible, generates export lists (or verify against a generated manifest) to reduce maintenance cost.

PR #218, opened 2026-07-14
Bump python-semantic-release and support release from minor and patch branches

AlexDanDuna:

Allow releasing new minor and patch versions from branches following the format `1.x' and '1.2.x':
- The GitHub Action was modified to trigger on pushes and PRs targetting "*.x" branches
- In order to allow for matching such branches for Python Semantic Release, it was required to bump `python-semantic-release` version as v7 (which was used previously) did not allow for regex in the branch name. I bumped the version to the latest v10.6.1, given the fact that v10 and v9 did not really introduce breaking changes, and upgrading to v8 would require adaptations anyway:
    - the toml configuration for python-semantic-release to match the new branch configuration format of python semantic release, as indicated [here](https://github.com/python-semantic-release/python-semantic-release/blob/v10.6.1/docs/upgrading/08-upgrade.rst#multibranch-releases) and [here](https://github.com/python-semantic-release/python-semantic-release/blob/v10.6.1/docs/concepts/multibranch_releases.rst)
    - `semantic-release publish` no longer creates the GitHub release, nor publishes to PyPi. According to the docs (see [here](https://github.com/python-semantic-release/python-semantic-release/blob/v10.6.1/docs/upgrading/08-upgrade.rst#repurposing-of-version-and-publish-commands) and [here](https://github.com/python-semantic-release/python-semantic-release/blob/v10.6.1/docs/upgrading/08-upgrade.rst#python-semantic-release-github-action)), this now has to be replaced by running `semantic-release version` (which builds the packages, creates the GH release, and the tag), publishing the package to the GH release by running `semantic-release publish`, and publishing the package to PyPi (I used twine for this).

To ease the  release of patched versions in the future. The implementation is generic and supports all branches that have a version, ending in ".x", as their names.

I tested the modifications on my own fork, with my own PyPi account:
- verified that the pipeline triggers work on the patch branches
- verified that the pipeline successfully publishes a GH Release and a PyPi release, when run on the patch branches

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -49,7 +49,15 @@ jobs:
   release:
     runs-on: ubuntu-latest
     needs: [build-lint-unit-test, integration-test]
-    if: github.event_name == 'push' && (github.ref == 'refs/heads/master' || github.ref == 'refs/heads/main')
+    if: github.event_name == 'push' &&
+        (
+          github.ref == 'refs/heads/master' ||
+          github.ref == 'refs/heads/main' ||
+          (
+            startsWith(github.ref, 'refs/heads/') &&
+            endsWith(github.ref, '.x')
+          )
+        )
     steps:
       - uses: actions/checkout@v3
         with:
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -2,9 +2,9 @@ name: Python package
 
 on:
   push:
-    branches: [master, main]
+    branches: [master, main, '*.x']
   pull_request:
-    branches: [master, main]
+    branches: [master, main, '*.x']
 
 jobs:
   build-lint-unit-test:
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -77,6 +77,12 @@ jobs:
           git config --global user.name "github-actions"
           git config --global user.email "action@github.com"
           semantic-release version
+        env:
+          # To generate a new token use the following guide, providing access to the "repo" scope.
+          # https://docs.github.com/en/enterprise-server@3.4/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token
+          # Once you update the secret on the repo, ensure you update the master/main protected branch policy
+          # to allow your user permission to "bypass required pull requests".
+          GH_TOKEN: ${{secrets.GH_TOKEN}}
       - name: Publish package distributions to PyPI using Twine
         if: steps.release.outputs.released == 'true'
         run: |
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -71,10 +71,23 @@ jobs:
           cache: "poetry"
       - run: poetry install
       - name: Semantic release
+        id: release
         run: |
-          pip install python-semantic-release==7.34.6
+          pip install python-semantic-release==10.6.1
           git config --global user.name "github-actions"
           git config --global user.email "action@github.com"
+          semantic-release version
+      - name: Publish package distributions to PyPI using Twine
+        if: steps.release.outputs.released == 'true'
+        run: |
+          pip install twine==6.1.0
+          python -m twine upload dist/*
+        env:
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -6,7 +6,7 @@ Contributions to *nisystemlink-clients-python* are welcome from all!
 the canonical upstream repository hosted on
 [GitHub](https://github.com/ni/nisystemlink-clients-python/).
 
-*nisystemlink-clients-python* follows a pull-request model for development.  If
+*nisystemlink-clients-python* follows ax ypull-request model for development.  If
```
rbell517: Looks like an erroneous change?

.github/workflows/python-package.yml
```diff
@@ -49,7 +49,15 @@ jobs:
...
     runs-on: ubuntu-latest
     needs: [build-lint-unit-test, integration-test]
-    if: github.event_name == 'push' && (github.ref == 'refs/heads/master' || github.ref == 'refs/heads/main')
+    if: github.event_name == 'push' &&
+        (
+          github.ref == 'refs/heads/master' ||
+          github.ref == 'refs/heads/main' ||
+          (
+            startsWith(github.ref, 'refs/heads/') &&
+            endsWith(github.ref, '.x')
```
rbell517: If possible we should use a strict regex here rather than just ends with .x. If someone made a user branch that ends with .x we don't want to run this. For example `users/rbell/fix-bug-for-1.6.1` would pass this check.

CONTRIBUTING.md
```diff
@@ -35,6 +35,28 @@ To contribute to this project, it is recommended that you follow these steps:
 6. Send a GitHub Pull Request to the main repository's master branch. GitHub
    Pull Requests are the expected method of code collaboration on this project.
 
+## Patching older versions
+1. Contact the [repository owners](https://github.com/AlexDanDuna/nisystemlink-clients-python/blob/master/.github/CODEOWNERS) for approval and ask them to create a
```
rbell517: You should link to master on the upstream repo, not your fork.

rbell517 (approved):

CONTRIBUTING.md
AlexDanDuna: Ouch, sorry. Thanks for pointing that out.

CONTRIBUTING.md
AlexDanDuna: Definitely is :( I modified this to trigger a pipeline run at some point.. I wasn't careful I also made the change on this branch.

.github/workflows/python-package.yml
AlexDanDuna: I agree and I also thought of such branches, but I was unable to find something better (regex) that works in this if, sadly. Still, even if the if passes for a branch that has a name similar to the one you suggested, no release will happen, because the semantic release config (in the toml file) has a strict regex, which will not match for branches that don't follow that format.

EDIT: Actually, was the example meant to have a .x at the end :)? It doesn't in its current form. I just assumed it was a branch name ending in "x" when writing the reply above.

AlexDanDuna: A message such as this one shows up on a branch that doesn't match the groups indicated in the toml config:
image

pyproject.toml
```diff
@@ -84,10 +84,16 @@ markers = [
 exclude = ".*\\.pyi"
 
 [tool.semantic_release]
-branch        = "master"
 version_toml  = ["pyproject.toml:tool.poetry.version"]
 build_command = "poetry build"
 
+# Allow releasing from the `master` branch.
+[tool.semantic_release.branches.master]
+match = "^master$"
```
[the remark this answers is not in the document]
AlexDanDuna: This repository does not have a `main` branch.

.github/workflows/python-package.yml
rbell517: Oh yes good point, I forgot that we would need literally a `.x` placeholder in the branch name, which makes this seem less likely and the fallback of semantic-release catching it is good enough for me.

PR #222, opened 2026-07-15
fix: Set SystemLink Server certificate path in JupyterHttpConfiguration, on Windows

AlexDanDuna:

SystemLink Server 26Q3 restricts notebook executions (by default). Access to the `HttpConfigurations` folder is
no longer granted for notebooks running under the JupyterHub/NotebookExecution services. Since config files under
`HttpConfigurations` are no longer readable, creating client objects from this library, from SLS notebooks,
will default to using `JupyterHttpConfiguration`. However, this does not set a `cert_path`, so HTTPS
requests will fail.

When running SystemLink Server notebooks (Windows), we will therefore pass the Web Server CA certificate's path as the 
`cert_path`, if the file exists. This will allow clients created with the default configuration to use this CA certificate.
If the file does not exist (when the Web Server is configured in HTTP mode), do not pass it; an invalid `cert_path`
will lead to errors.

Tracked under:
https://dev.azure.com/ni/DevCentral/_workitems/edit/3961101
and
https://dev.azure.com/ni/DevCentral/_queries/edit/3944027

It will allow customers whose notebooks use this library and break after upgrading to SLS 26Q3, to quickly fix them up by having an Administrator bump the library to the patched version that we'll release once this PR is merged.

Manual testing in a SLS instance. Verified that this snippet
`from nisystemlink.clients.tag import TagManager
tag_manager = TagManager()
tag_manager.open("tagname")`
yields no certificate errors when:
- Web Server is configured in HTTP mode (no certificate exists under Certificates/http-server)
- Web Server is configured in HTTPS mode, with a self-signed certificate (certificate exists).

### DIFF
--- a/nisystemlink/clients/core/_jupyter_http_configuration.py
+++ b/nisystemlink/clients/core/_jupyter_http_configuration.py
@@ -6,6 +6,7 @@ import os
 import sys
 
 from nisystemlink.clients import core
+from nisystemlink.clients.core._internal._path_constants import PathConstants
 
 
 class JupyterHttpConfiguration(core.HttpConfiguration):
@@ -14,8 +15,10 @@ class JupyterHttpConfiguration(core.HttpConfiguration):
     _HTTP_URI_ENV_VAR = "SYSTEMLINK_HTTP_URI"
     _HTTP_API_KEY_ENV_VAR = "SYSTEMLINK_API_KEY"
     _SYSTEMLINK_SERVER_CERT_PATH = (
-        r"C:\ProgramData\National Instruments\Skyline"
-        r"\Certificates\http-server\http-server.cer"
+        PathConstants.application_data_directory
+        / "Certificates"
+        / "http-server"
+        / "http-server.cer"
     )
 
     def __init__(self) -> None:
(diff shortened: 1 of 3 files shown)


nisystemlink/clients/core/_jupyter_http_configuration.py
```diff
@@ -12,6 +13,10 @@ class JupyterHttpConfiguration(core.HttpConfiguration):
 
     _HTTP_URI_ENV_VAR = "SYSTEMLINK_HTTP_URI"
     _HTTP_API_KEY_ENV_VAR = "SYSTEMLINK_API_KEY"
+    _SYSTEMLINK_SERVER_CERT_PATH = (
```
spanglerco: Can we use `PathConstants.application_data_directory` like we do in [`HttpConfigurationManager`](https://github.com/ni/nisystemlink-clients-python/blob/6c3e6bef321050de5ef39ae37ecd31f5df2ae4e0/nisystemlink/clients/core/_http_configuration_manager.py#L161)? That'll also give you a `pathlib.Path` instead of a string.

tests/core/test_jupyter_http_configuration.py
```diff
@@ -0,0 +1,95 @@
...
+from nisystemlink.clients.core import JupyterHttpConfiguration
+
+
+_HTTP_URI_ENV_VAR = "SYSTEMLINK_HTTP_URI"
+_HTTP_API_KEY_ENV_VAR = "SYSTEMLINK_API_KEY"
+_SYSTEMLINK_SERVER_CERTIFICATE_PATH = (
+    r"C:\ProgramData\National Instruments\Skyline"
+    r"\Certificates\http-server\http-server.cer"
+)
+_SYSTEM_LINK_API_KEY_HEADER = "x-ni-api-key"
```
spanglerco: nit: We should be consistent in naming between `_SYSTEMLINK_SERVER_CERTIFICATE_PATH` and `_SYSTEM_LINK_API_KEY_HEADER`.

spanglerco (changes requested):

nisystemlink/clients/core/_jupyter_http_configuration.py
AlexDanDuna: Thank you for the suggestion. I applied the change. However, I believe I'll have to do additional mocking because of this, since that constant does not give the expected path on linux, and the tests are run under linux.

AlexDanDuna: Done. Refactored the code a bit to ease mocking. The code uses PathConstants now.

spanglerco (approved):

PR #228, opened 2026-09-09
feat: Add filterType property to work item and work item template APIs models

gokulprasanth-ni:

Add filterType property to resource models in work item and template APIs

This support makes it easier for the users to create work items with either `Linq` or `Lucene` resource filters


nisystemlink/clients/work_item/models/_resources_definition.py
```diff
@@ -1,8 +1,16 @@
+from enum import Enum
 from typing import List
 
 from nisystemlink.clients.core._uplink._json_model import JsonModel
 
 
+class FilterType(str, Enum):
+    """The type of filter used to select the resources for the work item."""
```
snehaa-ni, line 8:
```suggestion
    """The query language used to interpret resource filters in the reservation."""
```

nisystemlink/clients/work_item/models/_resources_definition.py
```diff
@@ -1,8 +1,16 @@
...
+
+    LINQ = "LINQ"
+    LUCENE = "LUCENE"
```
snehaa-ni, line 11:
```suggestion
    LUCENE = "LUCENE"
    LINQ = "LINQ"
```

Since Lucene is the default FilterType, we could place it first.

nisystemlink/clients/work_item/models/_resources_definition.py
snehaa-ni: Similarly update in other places.

examples/work_item/work_items.py
```diff
@@ -100,6 +101,7 @@
                 ],
                 filter='modelName = "cRIO-9045" && serialNumber = "01E82ED0"',
             ),
+            filter_type=FilterType.LINQ,
```
snehaa-ni, line 104:
```suggestion
            filter_type="LINQ",
```

This is more readable and reflect the usage, can we update it in all 4 places in examples?

Image

Image

examples/work_item/work_items.py
```diff
@@ -155,7 +157,16 @@
 if created_work_item_id is not None:
     update_work_items_request = UpdateWorkItemsRequest(
         work_items=[
-            UpdateWorkItemRequest(id=created_work_item_id, name="Updated work item")
+            UpdateWorkItemRequest(
+                id=created_work_item_id,
+                name="Updated work item",
+                resources=ResourcesDefinition(
+                    systems=SystemResourceDefinition(
+                        filter='properties.data["Lab"] = "Battery Pack Lab"',
```
snehaa-ni, line 165:
```suggestion
                        filter='properties.data["Location"] = "Lab 1"',
```

This filter is already set in the create request model. Are we using the update API to set it again here?

same applies for templates example as well.

tests/integration/work_item/test_work_item_client.py
```diff
@@ -193,8 +194,9 @@ class TestWorkItemClient:
                             target_location_id="location-001",
                         )
                     ],
-                    filter="os:linux AND arch:x64",
+                    filter='os = "linux" && arch = "x64"',
                 ),
+                filter_type=FilterType.LINQ,
```
snehaa-ni, line 199:
```suggestion
                filter_type="LINQ",
```

tests/integration/work_item/test_work_item_client.py
```diff
@@ -267,6 +270,16 @@ def test__create_work_item__returns_created_work_items(
         )
         assert delete_work_item_template_response is None
 
+    def test__create_work_item__returns_resources_with_filter_type(
```
snehaa-ni, line 273: In my opinion, field-level validations (like asserting the response contains expected `filter_type`) are already covered by the service that the client calls. Our integration tests should only verify that the client correctly wires the request/response and deserializes the fields and don't throw error.

Other integration clients typically use focused assertions rather than validating every field in every response, so having a separate filter-type-specific test for each API feels inconsistent.

We could add the `filter_type` assertion to the existing update work items and templates tests alone, since the LINQ → LUCENE transition verifies the mutation behavior, and remove all the dedicated tests.

snehaa-ni (approved):

nisystemlink/clients/work_item/models/_resources_definition.py
gokulprasanth-ni: Updated

nisystemlink/clients/work_item/models/_resources_definition.py
gokulprasanth-ni: Updated

examples/work_item/work_items.py
gokulprasanth-ni: If I update like you suggested it is throwing some pipeline error. So use `FilterType.LINQ` itself.

image

tests/integration/work_item/test_work_item_client.py
gokulprasanth-ni: It is throwing some pipeline failure if I update as you suggested. So going with `FilterType.LINQ` itself.

tests/integration/work_item/test_work_item_client.py
gokulprasanth-ni: Updated as you suggested

examples/work_item/work_items.py
gokulprasanth-ni: Updated

examples/work_item/work_item_templates.py
```diff
@@ -50,6 +51,7 @@
             fixtures=TemplateResourceDefinition(
                 filter='modelName = "cRIO-9045" && serialNumber = "01E82ED0"'
             ),
+            filter_type=FilterType.LINQ,
```
rbell517: We want templates going forward to prefer Lucene, right? Should the examples reflect that? Or asked another way, is there a reason to prefer LINQ for these examples?

rbell517 (approved):

examples/work_item/work_item_templates.py
gokulprasanth-ni: No specific reason to prefer `LINQ` Richard. I have updated the examples with `LUCENE` filterType.

